PR #6, opened 2015-04-20
Added WorkerTest model

karthiksenthil:


dmorina:
docstrings and field names to be updated

PR #73, opened 2015-05-19
Makes the doc clear

pemmasanikrishna:
Have added some more info so that user can install it easily.


neilthemathguy:
Thank you for the PR, could you please add the suggestions to [FAQs](http://crowdresearch.stanford.edu/w/index.php?title=FAQs).. Let's keep `README` as minimal and informative as possible.

PR #79, opened 2015-05-21
Search task

Soroosh129:
The commits for using Tasks instead of Modules are there but reverted.
For now, hyperlinks redirect using module ID to /task?ID={moduleID}.

### DIFF
--- a/staticfiles/templates/tasksearches/tasklistSearch.html
+++ b/staticfiles/templates/tasksearches/tasklistSearch.html
@@ -75,7 +75,6 @@ _neil -->
 		<table st-table="displayedCollection" st-safe-src="rowCollection" class="table table-striped">
 			<thead>
 			<tr>
-                <th st-sort="description">ID</th>
 				<th st-sort="description">Description</th>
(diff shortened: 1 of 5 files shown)


dmorina:
Modules are a group of same tasks (hits) do not confuse them, you can have a module with only one task if you need it

staticfiles/templates/tasksearches/tasklistSearch.html
```diff
@@ -89,11 +90,13 @@
 			</thead>
 			<tbody>
 			<tr ng-repeat="row in displayedCollection">
+                <td>{{row.id}}</td>
 				<td>{{row.description}}</td>
 				<td>{{row.keywords}}</td>
 				<td>{{row.module_timeout}}</td>
 				<td>{{row.repetition}}</td>
 				<td>{{row.price}}</td>
+                <td><a ng-href="/task?ID={{row.id}}">detail</a></td>
```
nistala: is this linked to an url endpoint?

PR #114, opened 2015-05-25
fixes #113: add migrations/ to .gitignore

### LINKED ISSUE #113: update .gitignore to ignore crowdsourcing/migrations/ dir
avoid committing ```crowdsourcing/migrations/``` to codebase by mistake (like ```git add .```)

arcolife:
- closes crowdresearch/crowdsource-platform#113

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -106,3 +106,4 @@ bower_components
 static/bower_components
 # PyCharm
 .idea/
+migrations/


dmorina:
There are reasons why you would want to revision migrations, if you wanna go back to a certain migration..

nistala:
Migrations should be in github - http://stackoverflow.com/questions/10083130/what-is-the-correct-way-to-deal-with-db-migration-while-using-south-django-and

Issue #83, opened 2015-05-22
Username generator


dmorina:
When registering the system will give the user a username, that will be firstname.lastname, but only if the username does not exist yet.
Write an algorithm which generates a username in case the default pattern is taken by another user. This is based on the first name and last name, so if username firstname.lastname already exists it would use firstname.lastname1 or something like this. The user would be able to change this later on his profile/settings.

DRI @dmorina @neilthemathguy

Soroosh129:
This idea comes from a lazy guy, but couldn’t the username be the email address?


On May 23, 2015 at 2:28:13 AM, Durim Morina (notifications@github.com) wrote:

When registering the system will give the user a username, that will be firstname.lastname, but only if the username does not exist yet.
Write an algorithm which generates a username in case the default pattern is taken by another user. This is based on the first name and last name, so if username firstname.lastname already exists it would use firstname.lastname1 or something like this. The user would be able to change this later on his profile/settings.

angelamayx:
Couldn't they pick their own usernames?

dmorina:
@Soroosh129 currently if the username is taken then the email is used as a username.
we want usernames because emails are not good to be used in url lookups

@angelamayx they could do that while registering but I think we should keep registration form as short as possible, they will be able to change it later anyways..

v1p:
@dmorina Purpose of username is only to give a an SEO-friendly url to the user for his/her profile ?

dmorina:
there are few benefits of using a custom username: multiple ways to login, friendly urls, you can browse other users easily by using a username

shirishgoyal:
Another reason for not using emails is that most of the workers or requesters do not want to share their emails on the public pages of platform so username is used mostly for display purposes.

PR #116, opened 2015-05-25
#83 Username generator

shirishgoyal:
For username generator, we follow below in same order
a) Assign username as firstname.lastname
b) If a) is not unique, we find next available username+id
c) If b) exceeds max length(30), we assign email
d) if c) exceeds max length(30), we assign a random string of 30 characters


dmorina:
Good job so far, let's improve it a bit, a) is fine, b) will fail when other users change their usernames manually, and please put this in a function for reuse

PR #119, opened 2015-05-25
Username generator #83

shirishgoyal:
- Fixed the conflict for generating username if username is modifiable later
- Separated the function to generate unique increasing value for a field of any model


dmorina:
Good job, looking much better than before.
Now, let's optimize your solution, avoid loops even though they don't run many times, you can get the maximum match and increase the id from there..
Optimize the performance avoid duplicate actions.

Bonus point fill the gaps left between ids.

Issue #36, opened 2015-05-10
Provision Django with Gunicorn and Nginx.


neilthemathguy:
@dmorina @nistala @neilthemathguy

nistala:
We can set this up, let's pick a branch and set development environment: https://devcenter.heroku.com/articles/github-integration

neilthemathguy:
@nistala Sounds great!

neilthemathguy:
About Travis CI let's go for it!

PR #133, opened 2015-05-25
Setup heroku

nistala:
start work on #36

### DIFF
--- a/.buildpacks
+++ b/.buildpacks
@@ -1,0 +1,4 @@
+https://github.com/heroku/heroku-buildpack-nodejs.git
+ 
+ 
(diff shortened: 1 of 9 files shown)


dmorina:
in case of the auth_user does not exist error; view the migrations 
` heroku run python manage.py showmigrations `
and apply the auth migration first 
` heroku run python manage.py migrate auth `
after that you can continue with 
` heroku run python manage.py makemigrations crowdsourcing `
` heroku run python manage.py makemigrations oauth2_provider `
` heroku run python manage.py migrate `

dmorina:
looking good

PR #145, opened 2015-05-26
Adding Myself to Contributors List

pemmasanikrishna:


nistala:
LGTM

dmorina:
would be nice if you can resolve the conflict, we can do that too but it will help you for future work and it will speed up the whole process, thank you :)

PR #149, opened 2015-05-27
updated serializers , viewsets and urls.py

asmitagupta:
Added few more classes to the serializers and also the related viewswets .


dmorina:
accept the CLA please

csp/urls.py
```diff
@@ -15,7 +17,16 @@
 router.register(r'api/project', ProjectViewSet)
 router.register(r'api/category', CategoryViewSet)
 router.register(r'api/module', ModuleViewSet)
-router.register(r'api/projectrequester', ProjectRequesterViewSet)
+router.register(r'api/projectrequester', ProjectRequesterViewSet)router.register(r'api/WorkerSkill', WorkerSkillViewSet)
```
dmorina: new line?

csp/urls.py
```diff
@@ -15,7 +17,16 @@
...
+router.register(r'api/Worker', WorkerViewSet)
```
dmorina: use lower case and separate words with  a dash

nistala:
@dmorina LGTM from my end, can you send your LGTM as well?

asmitagupta:
i have added one more class so thats it nothing else to add

PR #168, opened 2015-05-28
Adding myself to contributor's page

JoshPitkofsky:


staticfiles/templates/contributors/pitkofsky.html
```diff
@@ -0,0 +1,54 @@
...
+            <br />
+            <br />
+          </div>
+
+        </div>
+
+        </div>
+    </div>
+</body>
+</html>
```
neilthemathguy: routers and home html are missing. Also test the code in local environment and fix few errors associated with pitkofsky.html

PR #178, opened 2015-05-29
Create saviaga.html

saviaga:
Create a new template at /staticfiles/templates/contributors/saviaga.html


neilthemathguy:
see #179

PR #222, opened 2015-06-01
Shift to angular-ui-router

shirishgoyal:
- Shifted from angular-router to angular-ui-router for flexible component based views
- Contributors page to use route
- Add missing dependency


dmorina:
lgtm 
@nistala what do you say? makes it easy to do multiple nesting, but do we wanna move away from ngRoute now?

Issue #238, opened 2015-06-08
Set the naming rule for AngularJS controllers, services, and views


ryosuzuki:
We need to have naming convention in order to make it easier to find files as well as add new features.

The naming rules should basically be followed with these conventions.
- [AngularJS Style Guide](http://www.codestyle.co/Guidelines/angularjs)
- [Extended AngularJS Style Guide](https://github.com/johnpapa/angular-styleguide)
- [RESTful API Resource Naming](http://www.restapitutorial.com/lessons/restfulresourcenaming.html)

PR #239, opened 2015-06-08
Naming convention for AngularJS

ryosuzuki:
Related issue: #238


staticfiles/js/tasks/controllers/task.list.controller.js
```diff
@@ -0,0 +1,38 @@
+(function() {
+  'use strict';
+
+  angular.module('crowdsource.tasks.controllers', ['smart-table'])
+    .factory('Task', Task);
+  Task.$inject = ['$http'];
+
+  angular.controller('TaskNewController', TaskController);
```
nistala: should this be task list?

nistala:
Any reason in particular for this naming convention? With angular we won't need separate controllers such as search unless we want a new page for them.

dmorina:
best practice and easiest to maintain is to group things by functionality, currently we are doing well so I don't think we should change it, let's revisit this later after  a month or so.. this doesnt mean that people can use random structures though, it's very simple group things that go together, and if the code is being shared by multiple pages then put it in a shared container

neilthemathguy:
+1

ryosuzuki:
OK, I understand.
We should focus on the existing issues

PR #242, opened 2015-06-08
Requester Task Portfolio and Profile additions

aditimithal:
Data fetched from json and displayed in Requester Task Portfolio
Personal information block made in worker's and requester's profile
Added myself to contributors.


staticfiles/js/requester/controllers/requester.controller.js
```diff
@@ -0,0 +1,80 @@
...
+  RequesterTaskPortfolioController.$inject = ['$location', '$scope', 'Authentication'];
+
+  /**
+  * @namespace HomeController
+  */
+  function RequesterTaskPortfolioController($location, $scope, Authentication) {
+    var vm = this;
+
+    $scope.user = {
+      TaskCategories : [{
```
nistala: would <i>really</i> like this to come from the backend using an api endpoint as we have set up with other routes

staticfiles/js/requester/controllers/requester.controller.js
```diff
@@ -0,0 +1,80 @@
...
+          NoofTasks : 30,
+	  reviews: 22,
+	  stars :3
+        },
+        
+      ]
+      
+    };
+ $scope.range = new Array(5);
+$scope.convertToInt= function (value) {
```
nistala: why do u need this?

staticfiles/js/requester/controllers/requester.controller.js
aditimithal: The backend api for this hasn't been made yet(totally). Once that is done , i will add it surely. But, till then shall this be merged so that i can clone the updated repo ?

staticfiles/js/requester/controllers/requester.controller.js
aditimithal: This json data is being used to display on the task portfolio of the requester.
scope.range and convertToInt is used for creating the stars (in task portfolio).

aditimithal: This json data is being used to display on the task portfolio of the requester.
scope.range and convertToInt is used for creating the stars (in task portfolio).

nistala:
LGTM

PR #248, opened 2015-06-10
Oauth complete

nistala:
- Set up redirects for forbidden errors
- Set up services for each feature
- Abstract away $http from every service and create a master http service to globally handle auth and auth errors
- Create a mock route that demonstrates what happens when auth error occurs


crowdsourcing/viewsets/project.py
```diff
@@ -78,12 +78,17 @@ def destroy(self, request, *args, **kwargs):
...
+        return Response(None, status=status.HTTP_403_FORBIDDEN)
+
     def get_queryset(self):
         queryset = Module.objects.all()
         requesterid=self.request.query_params.get('requesterid',None)
         if requesterid is not None:
             queryset = Module.objects.all().filter(owner__id=requesterid)
-        return queryset
+        # return queryset
+        return Response(None, status=status.HTTP_403_FORBIDDEN)
```
dmorina: no need for this, DRF handles it, we just set the permission_classes=[IsAuthenticated]

dmorina:
this is good, but everyone would have to use this http service, a better idea is to use interceptors
see the integrated angular-oauth2 which implements this in a neat way.
the only missing part is post configuration of client id and client secret

crowdsourcing/viewsets/project.py
nistala: this was only a test

nistala:
@dmorina do you have an example that shows what you're thinking? I'd rather wrap $http than change the base service to enable oauth because we don't want <i>all</i> our requests to contain auth tokens.

staticfiles/js/http.service.js
```diff
@@ -0,0 +1,57 @@
...
+      var deferred = $q.defer();
+      Authentication.attachHeaderTokens(settings);
+
+      $http(settings).success(function (data,config) {
+        deferred.resolve(arguments);
+      }).error(function (data, status, headers, config) {
+        deferred.reject(arguments);
+        // Handle authorization error, redirect to login.
+        if (status === 403) {
+          Authentication.unauthenticate();
```
dmorina: if access_token has expired please ask for a refresh token first and then retry.. if refresh_token is invalid then redirect to /login

dmorina:
fix merge conflicts, overall lgtm (see comments), and we have to tell others to use this http service instead of $http if they wanna get the right responses :)

nistala:
@dmorina addressed comments

dmorina:
project.service.js you have chosen yours I believe, it reverts few recent changes

nistala:
@dmorina reverted back project.py

Issue #252, opened 2015-06-10
Task Monitoring: convergence and HTML prototypeing


ryosuzuki:
There are two parts to this milestone. First, work with folks in #micro-macro to synthesize these designs and, as a group, come up with a mockup that represents your collective opinion on how this should work.

Second, create an HTML interface that translates these ideas into an HTML page. We prefer if you use the CSS templates from our github repository, check with folks in #research-engineering if you'd like help. Show how this would work with both a microtask such as webpage classification (e.g., is this web page a restaurant homepage, or not?) and with a macrotask such as copyediting an essay. The page can be static, but it should be a faithful mockup of the proposal.

neilthemathguy:
Not sure what is the purpose of this

PR #246, opened 2015-06-09
Requester portfolio design

Macbull:
Added some information to design of requester portfolio ( #64 )
here is the snapshot :
requester-portfolio3

PR #260, opened 2015-06-12
html for worker task portfolio

radhikabhanu:


neilthemathguy:
Let's synch this up as #252

nistala:
@radhikabhanu can u post a screenshot?

dmorina:
good job guys, do you think you could move the data to angular for now and do an ng-repeat on your worker/home html so that it reduces the amount of code?, and use ng-class with condtions on your data

radhikabhanu:
Here's the screenshot (This layout will be changed, and Angularised)
screen shot 2015-06-11 at 5 46 32 pm

dmorina:
lgtm, @nistala @neilthemathguy any thoughts?

neilthemathguy:
It is great, @radhikabhanu and @karanrajpal  are working on making it consistent with the #246. Please keep this mockup too, we will use it in future.

Issue #66, opened 2015-05-18
Requester Task Portfolio:  Data model and REST API


neilthemathguy:
___________________________________________________________________________________
### Task 
> 1. Enhance existing data model to support a requester's task dashboard 
> 2. Build REST API, view sets to support a requester's task dashboard  
> 3. Add dummy data to Validate the data model   
> 4.  At present, once a requester visits the task portfolio page, he get information about the tasks he has carried out. Tasks have ratings and reviews associated with it. This page can be public i.e. available to other workers and requesters.

___________________________________________________________________________________

### Technologies 
Python, Django Framework, REST API, PostgreSQL, Understanding of ORM 
___________________________________________________________________________________

### Related Issues
>1. Acceptance Test #68
>2. HTML, CSS, Bootstrap #64 
>3. Angular JS  #65 
>4. Rest API Data Models #66   	
>5. Automated Testing #67  

___________________________________________________________________________________

### Assigned 
GROUP 1
@Macbull @neilthemathguy

GROUP 2

___________________________________________________________________________________
### DRI
@dmorina  @elsabakiu  @neilthemathguy 
___________________________________________________________________________________

### Files
`crowdsourcing/models.py`
`crowdsourcing/serializers/requester.py`
`crowdsourcing/viewsets/requester.py`
`crowdsourcing/views.py`
`crowdsourcing/csp/urls.py`
`crowdsourcing/fixtures/`

___________________________________________________________________________________
### Reference
Below is an example of required information. This will change and evolve as we progress. Therefore, be creative and add further enhancements. Also check related issues to satisfy front end requirements

![Design Reference]
(http://crowdresearch.stanford.edu/w/img_auth.php/b/bf/Task_portfolio.png)

![Design Reference]
(http://crowdresearch.stanford.edu/w/img_auth.php/2/26/Data_Model_600_300ppi.jpg)

Macbull:
Can anybody please explain the meaning of Task, Project, Module and Template with example ?

neilthemathguy:
yes I will do it.

dmorina:
A project is the biggest unit say it's name is "Building a crowd sourcing platform", a module can be a part of the project say there are 2 modules "front end design", "database design". If they are left like this then these two have 1 task each.
If it's a microtask project like on AMT then a module would be a group of same kind Tasks(hits). Templates are to be defined by the requester when creating the module for a microtask project

Macbull:
thanks @dmorina

[...]

Issue #65, opened 2015-05-18
Requester Task Portfolio Angular


neilthemathguy:
### Task:  
> 1. Write Angular controllers and services to get the data for task portfolio
> 2. Write Angular controllers and services to get, update data from requester and task profile 
> 3. Use the API exposed by backend.  
> 4. Talk to backend folks to get the API link
> 5. Add ranking module (ui.bootstrap.rating) 
> 6. Post ratings via backend API  
>7. At present, once a requester visits the task portfolio page, he get information about the tasks he has carried out. Tasks have ratings and reviews associated with it. This page can be public i.e. available to other workers and requesters.
>8. Note* If backend is in progress then create dummy JSON, host it on the crowd research wiki and make a call to get the data 

___________________________________________________________________________________

### Technologies 
Angular JS, Java Script, HTML, CSS, Bootstrap
___________________________________________________________________________________

### Related Issues
>1. Acceptance Test #68
>2. HTML, CSS, Bootstrap #64 
>3. Angular JS  #65 
>4. Rest API Data Models #66   	
>5. Automated Testing #67  

___________________________________________________________________________________

### Assigned 
GROUP 1
 @neilthemathguy @aditimithal 

GROUP 2

___________________________________________________________________________________
### DRI
@neilthemathguy  @nistala 

___________________________________________________________________________________
### Reference
Below is an example of required information. This will change and evolve as we progress. Therefore, be creative and add further enhancements.  

![Design Reference]
(http://crowdresearch.stanford.edu/w/img_auth.php/b/bf/Task_portfolio.png)

Issue #64, opened 2015-05-18
Requester Task Portfolio Dashboard Design


neilthemathguy:
### Task 
> 1. Enhance existing requester task portfolio dashboard design and theme   
> 2. Build requester's task portfolio page using HTML, Bootstrap, CSS. Incorporate angular tags for integration with services and controllers. At present, once a requester visits the task portfolio page, he get information about the tasks he has carried out. Tasks have ratings and reviews associated with it.  This page can be public i.e. available to other workers and requesters.  

___________________________________________________________________________________

### Technologies 
HTML, CSS, Bootstrap, Angular JS (beginner)     
___________________________________________________________________________________

### Related Issues
>1. Acceptance Test #68
>2. HTML, CSS, Bootstrap #64 
>3. Angular JS  #65 
>4. Rest API Data Models #66   	
>5. Automated Testing #67  

___________________________________________________________________________________
### References & Resources 

[Bootstrap Getting Started](http://getbootstrap.com/getting-started/)
[Bootstrap Tutorials](http://www.w3schools.com/bootstrap/)
[Microsoft EdX Bootstrap] (https://www.edx.org/school/microsoft)
[Codecademy HTML, CSS] (http://www.codecademy.com/tracks/web) 
[Codeschool  HTML, CSS] (https://www.codeschool.com/paths/html-css)
 
___________________________________________________________________________________

### Assigned 
GROUP 1
 
GROUP 2

___________________________________________________________________________________
### DRI
@neilthemathguy  

___________________________________________________________________________________

### Current profile page
Below is an example of required information. This will change and evolve as we progress. Therefore, be creative and add further enhancements. Also check related issues to satisfy front end requirements

![Design Reference]
(http://crowdresearch.stanford.edu/w/img_auth.php/b/bf/Task_portfolio.png)

Macbull:
Shouldn't it be a Project or Module portfolio ? Only tasks are being referred throughout the current GUI . I am thinking of something like this. where rating is calculated by averaging all rating of task/modules. 
![requester dashboard](https://cloud.githubusercontent.com/assets/11361002/7903810/5f42bff0-0805-11e5-8d09-90351115992a.jpg)

dmorina:
yes, modules or projects should be used, if it's a macro-project then a task=module

Issue #35, opened 2015-05-10
Data Model Validation  


neilthemathguy:
This is a great task for someone who is looking to get familiar with existing data models. 

1) Explore current [data models](http://crowdresearch.stanford.edu/w/index.php?title=Data_Model_Creation)
2) Go through the code
3) Create JSON data files for each model. File will contain 5-10 records. 
4) If you think Data Models need an enhancement, please create an github issue include tags Need to be Prioritize, Data Models REST API) 
5) For more details and examples, please see [taskRankingData](https://github.com/crowdresearch/crowdsource-platform/blob/develop2/fixtures/neilTaskProfileData.json) or [taskRankingData](https://github.com/crowdresearch/crowdsource-platform/blob/develop2/fixtures/neilTaskRankingData.json)

PR #262, opened 2015-06-13
Requester portfolio backend 

Macbull:
Requester portfolio : data from get using /api/module/?portfolio=x where x is requester id(pk). 
-Some changes to statuses of different models.
-adding fields to module serializer
-slight modification to viewset
-some modifications to front-end
-updated json file to support current front-end
#66 #35 #64 #65 

Preview : (for testing purpose, i have explicitly set status of modules to 'finished', otherwise, only completed task will be shown in portfolio.)
requester-portfolio with backend


staticfiles/templates/requester/home.html
```diff
@@ -182,6 +182,46 @@ <h1 class="page-header">
 		    </div>
 		  </div>    
 	</div> 
+<<<<<<< HEAD
```
nistala: merge markers

crowdsourcing/models.py
```diff
@@ -168,18 +168,19 @@ class Module(models.Model):
...
     statuses = ((1, "Created"),
-                (2, 'In Progress'),
-                (3, 'In Review'),
+                (2, 'In Review'),
+                (3, 'In Progress'),
                 (4, 'Finished')
     )
     status = models.IntegerField(choices=statuses, default=1)
-    price = models.FloatField()
+    # price = models.FloatField()
```
dmorina: reason?

crowdsourcing/models.py
```diff
@@ -250,12 +252,11 @@ class TaskWorker(models.Model):
 
 class TaskWorkerResult(models.Model):
     task_worker = models.ForeignKey(TaskWorker)
-    template_item = models.ForeignKey(TemplateItem)
+    template_item = models.ForeignKey(TemplateItem,null = True)
```
dmorina: cannot be null

crowdsourcing/serializers/project.py
```diff
@@ -29,12 +31,12 @@ class ProjectSerializer(serializers.ModelSerializer):
 
     deleted = serializers.BooleanField(read_only=True)
     categories = serializers.PrimaryKeyRelatedField(queryset=models.Category.objects.all(), many=True)
-    start_date = serializers.DateTimeField()
-    end_date = serializers.DateTimeField()
+    start_date = serializers.DateTimeField(read_only=True)
```
dmorina: they must be editable

crowdsourcing/viewsets/project.py
```diff
@@ -89,9 +89,9 @@ class ModuleViewSet(viewsets.ModelViewSet):
 
     def get_queryset(self):
         queryset = Module.objects.all()
-        requesterid=self.request.query_params.get('requesterid',None)
```
dmorina: this goes by the request.user, not a query param

staticfiles/js/requester/controllers/requester.controller.js
```diff
@@ -25,15 +25,11 @@
...
-      var accepted = parseInt(parseInt(repetitions)/parseInt(totalSubmissions) *100).toFixed(2);
-            return accepted;
 
-        };
 
-	$scope.getRejectedRate= function (repetitions,totalSubmissions) {
-      var rejected = 100-(parseInt(parseInt(repetitions)/parseInt(totalSubmissions) * 100).toFixed(2));
-            return rejected;
+	$scope.getRate= function (num,totalSubmissions) {
+      var rate = (parseInt(parseInt(num)/parseInt(totalSubmissions) * 100).toFixed(2));
```
dmorina: i'd suggest to simplify this, not much parsing needed

crowdsourcing/viewsets/project.py
Macbull: but I think portfolio would be public, i.e one can see portfolio of others. !

dmorina: ah ok, then the lookup_field is id, use get_object it will automatically give you the correct requester

staticfiles/templates/requester/home.html
```diff
@@ -201,7 +239,7 @@ <h1 class="page-header">
 								<tbody>
 									<tr>
```
dmorina: indentation_

crowdsourcing/viewsets/project.py
Macbull: but, won't it return the requester in  Response ? , I want requester object to filter out the modules ! 
Plus, I was thinking of keeping this viewset more general, like, it can also be used to show portfolio of a worker also, by sending request /api/modle/?workerportfolio=3 or something like that !

dmorina: ok this is the ModuleViewset and it will be used in the main feed to show the available tasks out there where people can start working on them, there is mapping between tasks and workers and that's a different serialier/viewset where we query the tasks by worker id, we can query the projects by requester id and collect the existing modules of that project

Macbull: so should I make separate viewset and serializer for requester portfolio ?

dmorina: No, it should be available under the requester serializer

Macbull: You mean portfolio as a field in requester serializer, right ?

dmorina: Yeah

Macbull: ok ! but don't you think it will cause extra data transmission, while showing list of requesters ? 
Is it possible to select fields in serializer based on request ?

dmorina: This will be on a specific method and yes serializers can be dynamic

Macbull: Got it ! Thanks ! Will update the PR soon !

PR #268, opened 2015-06-14
Requester portfolio backend  

Macbull:
Some improvements to PR #262 .

1. Some of the serialisers are made dynamic, i.e. they now also accepts 'fields' argument. So that only relevant fields are included in response.
2. Now, portfolio will be requested from requester viewset, instead of module viewset .

#66 #65 #64


crowdsourcing/serializers/dynamic.py
```diff
@@ -0,0 +1,20 @@
...
+        fields = kwargs.pop('fields', None)
+
+        # Instantiate the superclass normally
+        super(DynamicFieldsModelSerializer, self).__init__(*args, **kwargs)
+
+        if fields is not None:
+            # Drop any fields that are not specified in the `fields` argument.
+            allowed = set(fields)
+            existing = set(self.fields.keys())
+            for field_name in existing - allowed:
```
dmorina: this is not used properly, you must pass the fields to the Serializer initializer

crowdsourcing/serializers/project.py
```diff
@@ -3,12 +3,15 @@
 from datetime import datetime
 from rest_framework import serializers
 from django.db.models import Avg
+from django.db.models import Max
+from django.db.models import Min
 from django.contrib.auth.models import User
 from django.core.exceptions import ObjectDoesNotExist
+from dynamic import DynamicFieldsModelSerializer
```
dmorina: do a module import here, this is marked as an invalid import

crowdsourcing/viewsets/requester.py
```diff
@@ -1,12 +1,22 @@
...
+from crowdsourcing.models import Module
 class RequesterViewSet(viewsets.ModelViewSet):
     from crowdsourcing.models import Requester
     queryset = Requester.objects.all()
     serializer_class = RequesterSerializer
+    @detail_route(methods = ['GET'])
+    def portfolio(self, request, *args, **kwargs):
+        requester = self.get_object()
+        modules = requester.module_set.all().filter(deleted=False,status=4)
+        serializer = ModuleSerializer(modules,many = True,fields = ('id', 'name', 'icon', 'project', 'categories',
```
dmorina: use either name params or normal ones, not a mix

crowdsourcing/serializers/dynamic.py
Macbull: It is getting fields from kwargs, I did't modified it, because we can modify it further to make it more dynamic for future needs.

crowdsourcing/serializers/project.py
Macbull: Sorry, I am not able to understand what you are saying.

crowdsourcing/serializers/dynamic.py
dmorina: alright

dmorina:
what's the module icon for?

Macbull:
icons for front end of portfolio. For example, fa-pencil for writing task.

dmorina:
those will be attached by task categories

Macbull:
but a module has multiple categories !

crowdsourcing/viewsets/requester.py
Macbull: not getting it ! any example ?

dmorina:
yeah, that's how we get the icons

crowdsourcing/viewsets/requester.py
dmorina: ModuleSerializer(modules,many = True,fields = ....
here your first parameter is not named but the rest are

crowdsourcing/serializers/project.py
dmorina: do this instead 
from crowdsourcing.serializers import dynamic

Macbull:
but front-end shows only one icon ! which one to show ?

dmorina:
front end shows icons for task categories, not for modules..

Macbull:
I am completely confused. 
Acc. to my current understanding, portfolio shows cards (or divisions) for each module of the requester, which may have multiple tasks and multiple categories. So we will need one icon for each module.
Please let me know, where I am getting it wrong. 
requester-portfolio with backend

crowdsourcing/viewsets/requester.py
Macbull: I am not able to find a way to send 'modules' as named parameter, neither to make 'many' as un-named ! Any ideas ?

dmorina: instance

Macbull: ohh! didn't expected it to work for many = true ! :D fixed it ! thanks

PR #273, opened 2015-06-19
Adding myself to the contributor's page

BUPTCS:
I have some problem in solving conflicts.
Of course I try my best to solve them.
I think it would be helpful to have your advice.
And I have done some assignments.

By the way, the github tell me I can automatically merge the pull request. However, the github also tell me that I could not create a pull request!


dmorina:
why so many changes for everyone else?
please get the most recent version and add your changes only, update only the parts where you need to add your code and not others'

Issue #76, opened 2015-05-20
Task Creation Micro Task Dynamic Templates: Angular


neilthemathguy:
__________________________________________________________________________________
### Task 
> 1. Write Angular controllers and services to get the data for generating dynamic templates and supporting the Task Creation workflow
> 2. Use the API exposed by backend.  
> 3. Talk to backend folks to get the API link
> 4. For Micro Task Dynamic Templates: You can use the `$compile` service to produce linker and then bind the template with a specific scope

### Task Creation Workflow that will need controllers and services
**Progress Headers and Forms Sequence**
> 1. Category Selection
> 2. Create task  
>3. Milestones or Input transducers (Input and output transducers: Tasks get vetted or improved by people on the platform immediately after getting submitted, and before workers are exposed to them, see the [research foundation] (http://crowdresearch.stanford.edu/w/index.php?title=GettingStarted)
>4. Task scheduler 
>5. Payment Mechanism  
>6. Task Review & Submit 

Be creative and enhance the suggested workflow to make it better 
___________________________________________________________________________________
**Form 1: Select the category** 
Progress HEADER: Highlight the Six Step Registration process, *Category Selection* Active others inactive 
> 1. Select the task from different categories
>2. User hits Next 
___________________________________________________________________________________

**Form 2: Dynamic Loading of the templates** 
Progress HEADER:  *Create Task* Active others inactive 
> 1. Present various templates thumbnails 
> 2. Let user select one of the templates  
>4. User hits Next 
___________________________________________________________________________________

**Form 3: Add Milestones, Rubrics, or Input Transducers** 
Progress HEADER *Add Milestone* Active others inactive 
> 1. Let user create the milestones to break down the project in multiple phases. Example: Milestone 0: Task Testing / Input Transducer 

___________________________________________________________________________________

**Form 4:  Task Scheduler**  
Progress HEADER *Task Scheduler* Active others inactive 
>1. Add calendar and time to indicate the task activation event
>2. Add the duration - how long the task will run? 
>3. (Ignore: Task Update: How can requester add more time to the existing task?)

___________________________________________________________________________________

**Form 5:  Payment Mechanism**
Progress HEADER  *Payment* Active others inactive 
>1. Current balance indicator
>2. Highlight minimum payment requirements 
>3. Interface to add more funds to support  

___________________________________________________________________________________

**Form 6:  Task Review & Submit**
Progress HEADER *Task Review* Active others inactive 
[...]

PR #276, opened 2015-06-25
Template builder for create task

nistala:
Preliminary work for template builder for create task.
- Integrate form builder (https://github.com/Selmanh/angularjs-form-builder)
- Merge into UI and integrate design
- Get directives and preview working

Pending
- Send data to backend
- Tie into create task workflow
- Show stored templates in "viewtemplate"

#76 

screenshot from 2015-06-25 03 15 30
screenshot from 2015-06-25 03 15 48


dmorina:
We should improve this with drag and drop and make it look aligned

nistala:
I think drag and drop will be a bit of work but let's make it look better!

Macbull:
May be this can help in drag and drop , https://github.com/Macbull/Bootstrap-Form-Builder

PR #281, opened 2015-06-29
Worker portfolio back end integration

Macbull:
Right now, models does not relate a module to any skill or skill level, so that is left as constant.

### DIFF
--- a/staticfiles/js/worker/services/worker.service.js
+++ b/staticfiles/js/worker/services/worker.service.js
@@ -29,16 +29,16 @@
         url: 'http://share-quick.com/cr/getWorkerProfile.php',
         method: 'POST'
       });
-      
-    }   
+
+    }
 
     function getWorkerTaskPortfolio() {
       return $http({
-        url: 'https://api.myjson.com/bins/q7jc',
+        url: '/api/worker/1/portfolio/',
         method: 'GET'
       });
-   }         
+   }
 
-    
+
   }
 })();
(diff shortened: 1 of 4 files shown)


dmorina:
They do, via module qualifications

Macbull:
qualification model is connected to module, but skill model is only connected to Worker via workerskills

dmorina:
Yeah and a qualification attribute can be a skill, we will cover this later don't worry about it

staticfiles/js/worker/services/worker.service.js
```diff
@@ -29,16 +29,16 @@
...
-    }   
+
+    }
 
     function getWorkerTaskPortfolio() {
       return $http({
-        url: 'https://api.myjson.com/bins/q7jc',
+        url: '/api/worker/1/portfolio/',
         method: 'GET'
       });
```
nistala: use HttpService when you're calling the api

Macbull: I tried that, but the whole page goes blank, after using that . Getting 304 response code !

nistala: you're not sending auth headers, so this won't work once auth is enabled.
304 means not modified, so something is wrong with the implementation

Macbull: I don't know what's going wrong ? HttpService doesn't seems to work with detail routes .

dmorina: httpservice should work just fine, the backend is not dependent on that

Macbull: The problem is solved , if I put 'return' before $http(settings).success(function (data,config) {
        deferred.resolve(arguments); in http.service.js

Macbull: the response code 304 was for static files, not for api request, sorry for confusion. But the problem still exist, I think HttpService is not returning response.

dmorina:
resolve merge conflicts and split this into two PRs, please do only one thing in one PR, it makes it easier to check and merge and that's how it's supposed to be :)

PR #283, opened 2015-06-29
Adding Permissions for Worker REST API

agarwal22megha:


dmorina:
Let's not have repeated code, IsOwnerOrReadOnly can go to a shared module which will be used by any viewset, here we keep only worker related permissions :)

PR #286, opened 2015-06-30
Tie in backend and frontend, add worker skill edit screen

nistala:


crowdsourcing/viewsets/worker.py
```diff
@@ -45,10 +45,12 @@ def destroy(self, request, *args, **kwargs):
 class WorkerViewSet(viewsets.ModelViewSet):
     queryset = Worker.objects.all()
     serializer_class = WorkerSerializer
+    lookup_field = 'profile'
     permission_classes = [IsOwnerOrReadOnly]
 
     @detail_route(methods=['post'], permission_classes=[IsAuthenticated])
     def update_worker(self, request, pk=None):
+        print request
```
dmorina: let's leave these out :)

staticfiles/templates/worker/account-change-password.html
```diff
@@ -0,0 +1,15 @@
+<div class="container bs-docs-container">
+  <div class="row">
+
+    <div class="col-md-3">
+      <ng-include src="'/static/templates/worker/account-navbar.html'"></ng-include>
+    </div>
+    <div class="col-md-9">
+
```
dmorina: this should go under user because it will be the same for worker and requester

PR #291, opened 2015-07-02
fixing bug

Macbull:
Worker model don't have owner attribute. we need a new permission for this.

### DIFF
--- a/crowdsourcing/viewsets/worker.py
+++ b/crowdsourcing/viewsets/worker.py
@@ -47,7 +47,7 @@ class WorkerViewSet(viewsets.ModelViewSet):
     queryset = Worker.objects.all()
     serializer_class = WorkerSerializer
     lookup_field = 'profile'
-    permission_classes = [IsOwnerOrReadOnly]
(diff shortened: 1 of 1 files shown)


crowdsourcing/viewsets/worker.py
```diff
@@ -47,7 +47,7 @@ class WorkerViewSet(viewsets.ModelViewSet):
     queryset = Worker.objects.all()
     serializer_class = WorkerSerializer
     lookup_field = 'profile'
-    permission_classes = [IsOwnerOrReadOnly]
+    # permission_classes = [IsOwnerOrReadOnly]
```
dmorina: because?

Macbull: Worker model does not have any 'owner' attribute

dmorina: ok, i see, then either this permission class needs to be updated OR the user can only update his own data

dmorina: merging for now

PR #293, opened 2015-07-02
Requester profile 

Macbull:
Adding some fields to serializer for requester profile


crowdsourcing/models.py
```diff
@@ -140,9 +140,9 @@ class Project(models.Model):
     name = models.CharField(max_length=128, error_messages={'required': "Please enter the project name!"})
     start_date = models.DateTimeField(auto_now_add=True, auto_now=False)
     end_date = models.DateTimeField(auto_now_add=True, auto_now=False)
-    owner = models.ForeignKey(Requester, related_name='project_owner')
+    owner = models.ForeignKey(Requester, related_name='projects_owned')
```
dmorina: wrong related name

crowdsourcing/models.py
```diff
@@ -140,9 +140,9 @@ class Project(models.Model):
...
     description = models.CharField(max_length=1024, default='')
-    collaborators = models.ManyToManyField(Requester, through='ProjectRequester')
+    collaborators = models.ManyToManyField(Requester, through='ProjectRequester', related_name='projects_collaborated')
```
dmorina: collaborators :)

crowdsourcing/serializers/requester.py
```diff
@@ -2,10 +2,26 @@
 
 from crowdsourcing import models
 from rest_framework import serializers
+import datetime
 
 class RequesterSerializer(serializers.ModelSerializer):
+    total_projects = serializers.SerializerMethodField()
+    current_projects = serializers.SerializerMethodField()
+    waiting_projects = serializers.SerializerMethodField()
+
```
dmorina: these functions are supposed to live under module

crowdsourcing/models.py
Macbull: I just thought tom make them more intuitive to use while using back-relations

crowdsourcing/serializers/requester.py
Macbull: but then we will need to pass extra parameter to module serializer ! 
PS : inspired from Worker REST API

crowdsourcing/models.py
dmorina: usually it's the related models combined

crowdsourcing/serializers/requester.py
dmorina: yeah, eventually worker would have to be updated to, let's not break the concept, it's ok to have a method in the module viewset which gets by requester or by worker

crowdsourcing/models.py
Macbull: oh! ok ! I will update these soon !

crowdsourcing/serializers/requester.py
dmorina: A serializer mainly deals with it's own Model and additional ones if any

Macbull: but these are only counts specific to a requester,  not specific to any module or project

dmorina: it doesnt matter, what matters is the object type and these are a list of modules so they go under ModuleSerializer

PR #299, opened 2015-07-04
AccountModel REST API

agarwal22megha:
1. Adding utility to list drive contents in google_drive.py
2. Creating REST API to list user's google (dropbox remains) drive contents
3. Fixing bug in viewsets/workers.py


crowdsourcing/viewsets/google_drive.py
```diff
@@ -76,4 +78,31 @@ def auth_end(self, request):
...
+
+class GoogleDriveUtil(APIView):
+    '''
+    Reference: https://developers.google.com/drive/v2/reference/children/list
+    Creates a Google Drive API service object and outputs the names and IDs
+    for files in google drive.
+    '''
+    def list_DriveContents(credentials):
+        credentials = Credentials.new_from_json(credentials)
+        http = credentials.authorize(httplib2.Http())
```
dmorina: build moves to Class init so that we can use it in other methods, we always need it

crowdsourcing/serializers/accountModel.py
```diff
@@ -0,0 +1,24 @@
...
+from crowdsourcing.viewsets.google_drive import GoogleDriveUtil
+
+class AccountModelSerializer (serializers.ModelSerializer):
+    drive_contents = serializers.SerializerMethodField()
+
+    class Meta:
+        model = models.AccountModel
+        read_only_fields = ('drive_contents')
+
+    def get_drive_contents(self, instance):
```
dmorina: it should accept a path, so that we will list the children

crowdsourcing/viewsets/google_drive.py
agarwal22megha: Added class __init__

crowdsourcing/serializers/accountModel.py
agarwal22megha: Modified to accept a folder id, and then shows all the children of that node

PR #304, opened 2015-07-06
fixed create userprofile api bug

lucamatsumoto:
I used the browser to test my fix at http://localhost:8000/api/profile/  
I also added unit test and test data but was unable to make it work because my test code did not properly load my test data. 
I included the data and unit test to show how I tested the code. 
Thank you :D let me know if you have questions.

### DIFF
--- a/crowdsourcing/models.py
+++ b/crowdsourcing/models.py
@@ -31,11 +31,17 @@ class Country(models.Model):
     created_timestamp = models.DateTimeField(auto_now_add=True, auto_now=False)
     last_updated = models.DateTimeField(auto_now_add=False, auto_now=True)
 
+    def __unicode__(self):
+        return u'%s' % (self.name)
+
 class City(models.Model):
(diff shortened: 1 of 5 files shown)


dmorina:
1. Address cannot be just an id
2. Username is the lookup_field so it wont be provided in the data itself

lucamatsumoto:
1. Are you suggesting including the entire address details in the create UserProfile payload, and then create a new Address entry at the same time?
I found that current /api/profile returns address as just id. That is why I assumed that address must be created separately via its own api up front. But if create should have the whole address details in the payload, should the apis to list and retrieve UserProfile show the details too?

2. Yes, username is for lookup, i.e. for retrieve and update. That is why I used the same for create api too for consistency. We need to point the auth user any way to create UserProfile. Should we use user id just for create api?

3. I also need comment out the test code since it is not working yet.

lucamatsumoto:
After chatting with Durim, we agreed on the payload and the strategy to fix this issue. I am closing this PR for now.

PR #309, opened 2015-07-07
Requester projects

Macbull:


crowdsourcing/viewsets/project.py
```diff
@@ -71,6 +71,12 @@ def list(self, request, *args, **kwargs):
         except:
             return Response([])
 
+    @list_route(permission_classes=[IsAuthenticated])
```
dmorina: methods=['get']

dmorina: remove permission_classes, it should be on Viewset level we have secured everything

PR #315, opened 2015-07-08
Enhancing GDrive Utilities

agarwal22megha:
1. Support file search by file name (relative to parent directory)


crowdsourcing/viewsets/google_drive.py
```diff
@@ -112,6 +112,13 @@ def list_files_in_folder(self, folder_id, q):
                 return message
         return file_list
 
+    def search_file(self, account_instance, file_title):
+         root_id = models.CredentialsModel.objects.get(account = account_instance).account.root
+         parentId = self.getPathId(root_id) #get the id of the parent folder
```
dmorina: you can fetch the root from accountmodel

PR #318, opened 2015-07-08
Task monitoring progression

aginzberg:
Integrated with my-projects and some front end work. It still needs some more front end work and testing - pretty close i think.


dmorina:
exclude migrations, and please pull latest changes, this modifies many things which is not supposed to happen

PR #326, opened 2015-07-09
homepage proposal claudia

saviaga:
homepage proposal Claudia

visit the homepage by going to http://localhost:8000/home


saviaga:
will open another, there is a file missing

PR #329, opened 2015-07-10
homepage proposal Claudia

saviaga:
Proposal for home page,  you visit the homepage by going to 
http://localhost:8000/home


dmorina:
we are not using FA, see material design icons
dont include the css here (put them in bower)
please follow the current developments, we are aiming for a clean simple design based in MD
is there a copyright for this image?

PR #330, opened 2015-07-12
Updates to Login Form, Signup Form, Footer, Brand name, Left Navbar, Time fields

shirishgoyal:
Handle error messages for Login and registration form
Add footer with contributors and social links
Add "Daemo" as brand name
Turn left navbar into responsive for different resolutions
Update time values to show as ago values
Wizard based design for Project Creation
Refactoring of functions for Wizard
Validations added for each step of wizard
Layout restructured for base template


nistala:
few things
- it's tough to review a PR with so many files, do we need 245 files for moment js? can u just check in the ones you're using?
- can u push this to an instance of heroku to review, I'm not sure what is needed to get it working.
- Are the changes to the layout based on suggestions from userresearch (if so, see point 1, it would be great to a demo for everyone to play with before checkin)

shirishgoyal:
1. Single PR with so many files - if PR is not pulled in quickly, all the new commits get added to same PR. If someone has a way to seperate them while PR doesn't get merged, please let me know. 
Moment JS - Personally I don't think bower libs should be part of git repository at all. I followed what was done before - See the commits to this folder before - https://github.com/crowdresearch/crowdsource-platform/tree/develop2/static/bower_components

2. Do we follow this approach of pushing to Heroku for all PRs?
Do DRIs review the commits by checkout before merging in?
No other DB/package level changes are needed for this PR apart from merging it. 

3.  All the changes in this PR are linked to issues with code, validation of fields and not layout modifications.

dmorina:
this PR has been reviewed long ago, at the submission time it had too many files too
the wizard layout breaks the UI principle, the idea is to keep it as simple as possible close to the Material Specs, there is no particular reason to limit the user to complete the information in any particular order (except in few questions)
do not add static files to git unless it's @nistala pushing to heroku

no need for social sites for now, contribs stay out as well

nistala:
@shirishgoyal 
1) We added bower components because our heroku deploys don't allow to install nodejs and python at the same time, the build pack ordering does nodejs first which doesn't build a python system - please refer to slack, there was a discussion about this. The problem is not with your commits but the sheer number of files that make it difficult for review.
2) Pushing to Heroku is outlined in the README - you don't need this PR to be merged to push to your heroku instance.
3) No, you've made many layout modifications, form fields including adding links that we haven't decided to add back(contributors).

PR #333, opened 2015-07-15
fixed create user profile api, added unit test

lucamatsumoto:
I fixed the user profile create api. Also, I added a unit test. If this is ok, I can fix the updates as well.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -74,7 +74,7 @@ Vagrant.configure(2) do |config|
      sudo apt-get update
      sudo apt-get install -y git nodejs nodejs-legacy npm python-pip python-dev postgresql-client postgresql postgresql-contrib postgresql-server-dev-all
      sudo npm install -g bower
-     sudo -u postgres psql postgres -c "CREATE USER vagrant;"
+     sudo -u postgres psql postgres -c "CREATE USER vagrant CREATEDB;"
      sudo -u postgres psql postgres -c "CREATE DATABASE crowdsource_dev ENCODING 'UTF8';"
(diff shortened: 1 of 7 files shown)


lucamatsumoto:
My unit test works fine on vagrant but not on Travis. Is this a database issue on Travis?

Vagrantfile
```diff
@@ -74,7 +74,7 @@ Vagrant.configure(2) do |config|
      sudo apt-get update
      sudo apt-get install -y git nodejs nodejs-legacy npm python-pip python-dev postgresql-client postgresql postgresql-contrib postgresql-server-dev-all
      sudo npm install -g bower
-     sudo -u postgres psql postgres -c "CREATE USER vagrant;"
```
dmorina: unnecessary change?

crowdsourcing/serializers/utils.py
```diff
@@ -9,21 +9,68 @@ class RegionSerializer(serializers.ModelSerializer):
...
+        return region
+
 
 class CountrySerializer(serializers.ModelSerializer):
+    region = RegionSerializer()
+
     class Meta:
         model = models.Country
 
+    def create(self, **kwargs):
```
dmorina: these are pre loaded by the system, the user will select from a dropdown or sth

Vagrantfile
lucamatsumoto: When I run unit test on Vagrant, it creates test_crowdsource_dev database by using vagrant account. Without adding CREATEDB, it didn't work.

PR #335, opened 2015-07-22
adding to contributor page

unPourToutEtToutPourUn:
not sure if i'm doing it correctly :( and how do i add a photo? 
image


dmorina:
only the route is not enough, we are not accepting these anymore.. we might add everyone at the end

PR #340, opened 2015-07-29
Task creation PR

nistala:
Get task creation working

http://lit-hollows-1533.herokuapp.com/


dmorina:
resolve conflicts first and then we will review it

static/templates/project/base.html
```diff
@@ -11,9 +11,9 @@
       <div class="project-form">
           <h3 class="step">{{project.getStepName(project.getStepId())}}</h3>
           <div class="step-content">
-              <ng-include ng-show="project.getStepId()==1" src="'/static/templates/project/categories.html'"></ng-include>
-              <ng-include ng-show="project.getStepId()==2" src="'/static/templates/project/details.html'"></ng-include>
-              <ng-include ng-show="project.getStepId()==3" src="'/static/templates/project/milestones.html'"></ng-include>
+              <ng-include ng-show="project.getStepId()==1" src="'/static/templates/project/gettingstarted.html'"></ng-include>
```
dmorina: no need to rename files, if we are already under /project we know that the details are of the project

crowdsourcing/serializers/project.py
```diff
@@ -36,12 +36,15 @@ class ModuleSerializer(DynamicFieldsModelSerializer):
     deleted = serializers.BooleanField(read_only=True)
     template = TemplateSerializer(many=True, read_only=False)
     module_tasks = TaskSerializer(many=True, read_only=True)
+    csv_data = serializers.CharField(read_only=True)
 
     def create(self, **kwargs):
         templates = self.validated_data.pop('template')
         project = self.validated_data.pop('project')
+        csv_data = self.initial_data.pop('csv_data')
```
ksetyadi: I'm not sure whether you have validate the `csv_data` because you get it right away from `initial_data`. CMIIW.

crowdsourcing/serializers/project.py
```diff
@@ -49,17 +52,19 @@ def create(self, **kwargs):
             for item in template_items:
                 models.TemplateItem.objects.get_or_create(template=t[0], **item)
         if module.has_data_set:
```
ksetyadi: Did the `if` and the `else` part misplaced?

If `module.has_data_set` should create tasks but the code (to create such tasks) is in the `else` part.

crowdsourcing/viewsets/google_drive.py
```diff
@@ -83,6 +85,67 @@ def auth_end(self, request):
...
+        prev = AccountModel.objects.get(owner=request.user, type='GOOGLEDRIVE').root
+        if parent != "":
+            account = 1
+            drive_util = GoogleDriveUtil(account_instance=account)
+            file_list = drive_util.list_files_in_folder(prev, "blah")
+            for fileobj in file_list:
+                if fileobj['title'] == parent:
+                    prev = fileobj['id']
+                    break
+        account = 1
```
ksetyadi: what is `account = 1` exactly means? If it's a constant, we can make it as a constant; else we shouldn't hard code it here. We can optionally put it as parameter if needed.

crowdsourcing/viewsets/google_drive.py
```diff
@@ -83,6 +85,67 @@ def auth_end(self, request):
...
+                curr_arr = []
+            else:
+                curr_string += row[0]
+        curr_arr.append(curr_string)
+        formatted_file.append(curr_arr)
+        return Response(formatted_file, 200)
+
+    def get_files(self, request):
+        parent = request.data['parent']
+        account = 1
```
ksetyadi: Please see https://github.com/crowdresearch/crowdsource-platform/pull/340/files#r35901096

bower.json
```diff
@@ -22,7 +22,9 @@
     "angular-oauth2": "3.0.1",
     "angular-filter": "~0.5.4",
     "angular-dragdrop": "1.0.9",
-    "angular-ui-sortable": "~0.13.4"
+    "angular-ui-sortable": "~0.13.4",
+    "ng-file-upload": "~5.0.9",
+    "mdi": "~1.1.70"
```
dmorina: we already have imported the font icons, in case you need the svgs then use material-design-icons

crowdsourcing/serializers/project.py
```diff
@@ -36,12 +36,15 @@ class ModuleSerializer(DynamicFieldsModelSerializer):
     deleted = serializers.BooleanField(read_only=True)
     template = TemplateSerializer(many=True, read_only=False)
     module_tasks = TaskSerializer(many=True, read_only=True)
+    csv_data = serializers.CharField(read_only=True)
```
dmorina: read_only means that this field will be serialized when reading (GET request) i.e it's not used in POST/PUT

crowdsourcing/serializers/project.py
dmorina: if you fix the other problem it should appear in validated_data

crowdsourcing/serializers/project.py
```diff
@@ -49,17 +52,19 @@ def create(self, **kwargs):
             for item in template_items:
                 models.TemplateItem.objects.get_or_create(template=t[0], **item)
         if module.has_data_set:
-            pass # spreadsheet or drive import
+            # create tasks here.
+            pass
         else:
-            task = {
```
dmorina: you need this if no data set is provided

crowdsourcing/serializers/project.py
```diff
@@ -97,6 +102,7 @@ class Meta:
                   'categories', 'modules')
 
     def create(self, **kwargs):
+        csv_data = self.initial_data['modules'][0]['csv_data']
```
dmorina: this is not necessary it's supposed to just be passed to the module serializer

crowdsourcing/viewsets/google_drive.py
```diff
@@ -83,6 +85,67 @@ def auth_end(self, request):
 class GoogleDriveViewSet(ViewSet):
     permission_classes = [IsAuthenticated]
 
+    def add_folder(self, request):
```
dmorina: misleading name, account is wrong

crowdsourcing/viewsets/google_drive.py
```diff
@@ -83,6 +85,67 @@ def auth_end(self, request):
...
+            for fileobj in file_list:
+                if fileobj['title'] == parent:
+                    prev = fileobj['id']
+                    break
+        account = 1
+        drive_util = GoogleDriveUtil(account_instance=account)
+        file = drive_util.create_folder(name, prev)
+        return Response({'id': file['id']}, 200)
+
+    def parse(self, request):
```
dmorina: does not belong here, it can be called from any other module so it goes elsewhere in utils

crowdsourcing/viewsets/google_drive.py
```diff
@@ -83,6 +85,67 @@ def auth_end(self, request):
...
+        uploadedFile = request.data['file']
+        csvinput = csv.DictReader(uploadedFile)
+        arr = []
+        for row in csvinput:
+            arr.append(row)
+        return Response(arr)
+
+    def parse_backup(self, request):
+        parent = request.data['parent']
+        account = 1
```
dmorina: account error here too

csp/urls.py
```diff
@@ -53,6 +53,9 @@
   url(r'^api/google-drive/init', GoogleDriveOauth.as_view({'post': 'auth_init'})),
   url(r'^api/google-drive/finish', GoogleDriveOauth.as_view({'post': 'auth_end'})),
   url(r'^api/google-drive/list-files', GoogleDriveViewSet.as_view({'get': 'query'})),
+  url(r'^api/google-drive/add-folder', GoogleDriveViewSet.as_view({'post': 'add_folder'})),
+  url(r'^api/google-drive/parse-csv', GoogleDriveViewSet.as_view({'post': 'parse'})),
+  url(r'^api/google-drive/get-files', GoogleDriveViewSet.as_view({'post': 'get_files'})),
```
dmorina: duplicate list-files is intended for this purpose

nistala:
@aginzberg you can remove the google drive stuff from this PR since this PR is not concerning that integration.

dmorina:
fix 404 on the base template

crowdsourcing/serializers/project.py
```diff
@@ -36,12 +36,17 @@ class ModuleSerializer(DynamicFieldsModelSerializer):
     deleted = serializers.BooleanField(read_only=True)
     template = TemplateSerializer(many=True, read_only=False)
     module_tasks = TaskSerializer(many=True, read_only=True)
+    csv_data = serializers.ListField(write_only=True)
```
dmorina: good

crowdsourcing/serializers/project.py
```diff
@@ -36,12 +36,17 @@ class ModuleSerializer(DynamicFieldsModelSerializer):
...
+    has_data_set = serializers.BooleanField(write_only=True)
```
dmorina: why is it write_only? isn't it both?

crowdsourcing/viewsets/csvmanager.py
```diff
@@ -0,0 +1,42 @@
...
+    #         if row == []:
+    #             curr_arr.append(curr_string)
+    #             formatted_file.append(curr_arr)
+    #             curr_string = ""
+    #             curr_arr = []
+    #         else:
+    #             curr_string += row[0]
+    #     curr_arr.append(curr_string)
+    #     formatted_file.append(curr_arr)
+    #     return Response(formatted_file, 200)
```
dmorina: remove commented out code

static/js/project/controllers/project.controller.js
```diff
@@ -194,10 +212,10 @@
       }
       function getStepName(stepId){
           if(stepId==1){
-              return '1. Category';
+              return '1. Getting Started';
```
dmorina: no sure about this, these are definitely categories so it can stay as is, let's worry about renaming later

static/js/project/controllers/project.controller.js
```diff
@@ -271,5 +293,68 @@
       function monitor(project) {
         window.location = 'monitor/' + project.id;
       }
+
+      function addDriveFolder(name) {
```
dmorina: remove unused function

static/js/project/controllers/project.controller.js
```diff
@@ -271,5 +293,68 @@
...
+              });
+          },
+          function error(resp) {
+            console.log("boooo");
+
+          }).finally(function () {
+
+          });
+      }
+      function getFiles() {
```
dmorina: this one too

static/js/project/services/project.service.js
```diff
@@ -119,5 +124,29 @@
       instance = {};
     }
 
+    function addDriveFolder(name, parent) {
```
dmorina: remove these unused functions

static/js/task-feed/controllers/task-feed.controller.js
```diff
@@ -27,8 +27,44 @@
       self.modules = [];
 
       TaskFeed.getProjects().then(
-        function success (data) {
-            self.modules = data[0];
+        function success (successData) {
```
dmorina: this one does not work, take the current version which fixes this (work in progress on task feed)

static/templates/base/index.html
```diff
@@ -11,6 +11,7 @@
     <title>CrowdResearch</title>
     <!-- Custom CSS -->
     <link href="{% static 'bower_components/angular-material/angular-material.min.css' %}" rel="stylesheet">
+    <link href="{% static 'bower_components/mdi/css/materialdesignicons.min.css' %}" rel="stylesheet" media="all" type="text/css" />
```
dmorina: 404

static/templates/project/categories.html
```diff
@@ -1,18 +0,0 @@
-<div layout="row">
```
dmorina: no need to rename these files

static/templates/project/prototypetask.html
```diff
@@ -0,0 +1,61 @@
+<md-subheader>
```
dmorina: this will be used for any milestone so renaming doesnt make sense, let's leave these as they are

static/js/task-feed/controllers/task-feed.controller.js
nistala: existing version picks the first module in the list, that's not correct.

dmorina: the checks have to happen in backend, this is not the way to do it in front end, either way the the logic of checking here is wrong

crowdsourcing/serializers/project.py
aginzberg: How can it be both read_only and write_only? We only want it when creating/updating the instance hence write_only?

dmorina: sorry if the comment was confusing I mean it's read and write, that means you dont pass any of the two parameters

aginzberg: sorry still a bit confused. when i dont pass either parameter i start getting an error from the GET request to requester_projects when it tries to access the serialized data. does that mean i have to add csv_data to the module model?

aginzberg: misread the comments. resolved.

static/js/task-feed/controllers/task-feed.controller.js
nistala: let's remove it when it's implemented and not wait for it. If you want it changed - let me know what logic is wrong it.

crowdsourcing/viewsets/csvmanager.py
```diff
@@ -0,0 +1,13 @@
...
+
+class CSVManagerViewSet(ViewSet):
+
+    def parse(self, request):
+        uploadedFile = request.data['file']
+        csvinput = csv.DictReader(uploadedFile)
+        arr = []
+        for row in csvinput:
+            arr.append(row)
+        return Response(arr)
```
dmorina: great that you moved this to it's own class so that we can reuse it
we should improve this because the data is doing a round trip and another POST that is: it's being uploaded, processed, returned back to the client and uploaded again (pretty much), you see the problem right?
an idea would be you upload the file once, process the metadata only (store the file, I can provide you a model for now) and return only the metadata which means column names and number of rows and a file id
now you are not returning the data anymore, and you are not resending the data to the server again because we already have it.
@ksetyadi  @neilthemathguy  your input please, thank you

crowdsourcing/models.py
```diff
@@ -467,4 +468,24 @@ class ConversationRecipient(models.Model):
...
-    deleted = models.BooleanField(default=False)
\ No newline at end of file
+    deleted = models.BooleanField(default=False)
+
+class File(models.Model):
+    #TODO will need save files on a server rather than in a temporary folder
+    file = models.FileField(upload_to='tmp/')
+    deleted = models.BooleanField(default=False)
+
+    def parse_csv(self):
```
nistala: these functions should be in the viewset as you had it before.

crowdsourcing/serializers/project.py
```diff
@@ -36,12 +35,20 @@ class ModuleSerializer(DynamicFieldsModelSerializer):
     deleted = serializers.BooleanField(read_only=True)
     template = TemplateSerializer(many=True, read_only=False)
     module_tasks = TaskSerializer(many=True, read_only=True)
+    file_id = serializers.IntegerField(write_only=True)
 
     def create(self, **kwargs):
         templates = self.validated_data.pop('template')
         project = self.validated_data.pop('project')
+        file_id = self.validated_data.pop('file_id')
```
nistala: is the id attribute created by the file serializer unique? what if i upload the same file twice, would it produce different ids?

aginzberg: yes it produces different ids

crowdsourcing/models.py
aginzberg: spoke to durim about this. consensus was that model methods was the way to go stylistically. because we aren't parsing/deleting the file until we are serializing the module i don't think it really makes sense to have these in the viewset.

crowdsourcing/models.py
```diff
@@ -467,4 +468,24 @@ class ConversationRecipient(models.Model):
 class UserMessage(models.Model):
     message = models.ForeignKey(Message)
     user = models.ForeignKey(User)
-    deleted = models.BooleanField(default=False)
\ No newline at end of file
+    deleted = models.BooleanField(default=False)
+
+class File(models.Model):
```
nistala: "File" seems like a common term and could be taken by other libraries, could we call it RequesterInputFile or InputCSVFile (then you can keep the parse function here)

aginzberg: Agreed. will resolve this after the others make comments.

crowdsourcing/serializers/project.py
nistala: great!

crowdsourcing/models.py
dmorina: as this is just processing the file content i.e it's representation it is supposed to be here, and it will be reused by the other backend classes, this is all good

crowdsourcing/viewsets/file.py
```diff
@@ -0,0 +1,23 @@
...
+		csvinput = csv.reader(uploadedFile)
+		column_headers = csvinput.next()
+		num_rows = sum(1 for row in csvinput)
+		file_serializer = FileSerializer(data=request.data)
+		if file_serializer.is_valid():
+			id = file_serializer.create()
+			metadata = {'id': id, 'num_rows': num_rows, 'column_headers': column_headers}
+			return Response({'metadata': metadata})
+		else:
+			return Response(file_serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
dmorina: minor thing: you probably want to rename this to sth else as it is actually saving the file too

crowdsourcing/serializers/file.py
```diff
@@ -0,0 +1,17 @@
+from crowdsourcing import models
+from rest_framework import serializers
+from crowdsourcing.models import File
+from crowdsourcing.serializers.file import File
+from crowdsourcing.serializers.dynamic import DynamicFieldsModelSerializer
+
+class FileSerializer(DynamicFieldsModelSerializer):
```
dmorina: hyperlinkedserializer would be the best here in case you need to fetch the file later but this will work for now since we dont need to download the file

PR #341, opened 2015-07-31
New Drive button and design task sidenav

rahulsheth1016:
1. I implemented new changes to the buttons on the milestones document.
I created a new button that allows for the user to bring files from
their Dropbox, which is another very popular drive.
2. I made new changes to the design task navbar.
    a. Added the sections: Project, Objective, Task. This should
clarify the navbar. One thing to think about moving forward is how to
include the radio buttons, checkboxes, etc. Perhaps they can be
versions of tasks?
    b. Added a feature where the user can click a button to have popup
where the user can read the description for a component.

### DIFF
--- a/static/js/template/controllers/template.controller.js
+++ b/static/js/template/controllers/template.controller.js
@@ -47,62 +47,77 @@
     self.templateComponents = [
       {
         id: 1,
-        name: "Label",
+        name: "Project",
         icon: null,
         type: 'label',
-        description: "Use for static text: labels, headings, paragraphs"
+        description: "Title of the Project. For example, if you were designing a website, this would be "Design a Website for me.""
+      },
+      { 
+	id: 2, 
+	name: "Objective", 
+	icon: null, 
+	type: 'label', 
+	description: "One task associated with the project. For example, if you were designing a website, an example objective could be "Create a navigation bar in the header."" 
       },
       {
-        id: 2,
+	id: 3, 
+	name: "Task",
+	icon: null, 
+	type: 'label', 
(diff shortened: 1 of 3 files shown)


static/js/template/controllers/template.controller.js
```diff
@@ -47,62 +47,77 @@
...
+      { 
+	id: 2, 
+	name: "Objective", 
+	icon: null, 
+	type: 'label', 
+	description: "One task associated with the project. For example, if you were designing a website, an example objective could be "Create a navigation bar in the header."" 
+      },
+      {
+	id: 3, 
+	name: "Task",
```
dmorina: this will confuse people, we will figure this out in a different way, if it's a macro project we will skip Task Design and just give them a text area

static/templates/project/milestones.html
```diff
@@ -44,6 +44,38 @@
 	        </md-input-container>
 		</div>   
 		</div>
+	<md-radio-button value="dropbox">Dropbox</md-radio-button>
```
dmorina: dropbox will come at a later stage

static/templates/template/components.html
```diff
@@ -5,7 +5,9 @@
                       jqyoui-draggable="{index: {{$index}}, animate: true, placeholder: 'keep'}" class="md-2-line" >
             <div class="md-list-item-text">
                 <h3>{{ component.name }}</h3>
-                <p>{{component.description}}</p>
+		<button type="button" class="btn btn-default" onclick="alert('{{component.description}}')">Click to see component description!</button>
```
dmorina: no alerts

PR #343, opened 2015-08-04
User profile

radhikabhanu:
1. Making a call to api/profile for user's profile (regardless of whether user is worker or req)
2. Small fix to the UserProfileViewSet


dmorina:
let's not add bower_components to git

radhikabhanu:
New controller and service for generic user profile
Contributor - Karan Rajpal

dmorina:
see comment above

nistala:
@radhikabhanu @karanrajpal please remove the bower_components from the PR and we can do the review after

PR #344, opened 2015-08-05
Monitoring Backend

aginzberg:
Added two filters. One to grab all of a requesters modules (essentially identical to Elsa's requester_projects) and one to grab the data needed for monitoring (tasks for the data field, task_workers for the worker alias, and task_worker_results for the result).

### DIFF
--- a/crowdsourcing/viewsets/worker.py
+++ b/crowdsourcing/viewsets/worker.py
@@ -5,12 +5,11 @@ from crowdsourcing.serializers.project import *
 from crowdsourcing.models import *
 from rest_framework import status, viewsets
 from rest_framework.response import Response
-from rest_framework.decorators import detail_route
+from rest_framework.decorators import detail_route, list_route
 from rest_framework.permissions import IsAuthenticated
 from django.shortcuts import get_object_or_404
 from crowdsourcing.permissions.util import *
 from crowdsourcing.permissions.user import IsWorker
-
 
 class SkillViewSet(viewsets.ModelViewSet):
     queryset = Skill.objects.all()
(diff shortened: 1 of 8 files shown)


crowdsourcing/viewsets/worker.py
```diff
@@ -156,6 +155,16 @@ def retrieve(self, request, *args, **kwargs):
         serializer = TaskWorkerResultSerializer(instance=worker)
         return Response(serializer.data)
 
+    @list_route(methods=['POST'])
```
dmorina: you mean GET right?

static/js/monitor/services/monitor.service.js
```diff
@@ -38,10 +38,13 @@
     }
 
 
-    function getTaskWorkerResults(moduleId){
+    function getMonitoringData(module_id){
       var settings = {
-        url: '/api/task/' + moduleId + '/',
-        method: 'GET'
+        url: '/api/task-worker-result/monitoring_data/',
+        method: 'POST',
```
dmorina: GET?

crowdsourcing/viewsets/worker.py
```diff
@@ -156,6 +155,16 @@ def retrieve(self, request, *args, **kwargs):
...
+    def monitoring_data(self, request, **kwargs):
+        module = Module.objects.get(id=request.data['module_id'])
```
dmorina: test this after changing to GET, it will be empty, params are stored elsewhere

elsabakiu:
@aginzberg cool, just keep in mind that project objects returned by the api(requester_projects) contain all the corresponding modules, the modules the corresponding tasks and so on. For example, you can get the modules as follows: modules = project.modules; (you can refer to the Monitor.getProject() function in monitor.controller.js for more examples) When possible, use this way to get the data rather then making a duplicate api call.

dmorina:
@elsabakiu we will drop the tasks from `/api/project/` (just task count is sufficient) these serializers are dynamic so we can choose on what call which fields we need

elsabakiu:
Yes, sounds even better. Send to the frontend only the data needed for a specific page + use related entities directly rather then making duplicate api calls :)

crowdsourcing/viewsets/worker.py
```diff
@@ -156,6 +155,13 @@ def retrieve(self, request, *args, **kwargs):
         serializer = TaskWorkerResultSerializer(instance=worker)
         return Response(serializer.data)
 
+    @list_route(methods=['GET'])
+    def monitoring_data(self, request, **kwargs):
```
dmorina: this goes to task viewset

PR #349, opened 2015-08-07
Basic monitor interface

aginzberg:
mostly frontend changes - with a few small backend things. feel free to checkout my fixture ginzbergMonitorData.json.

Demo - https://lit-hollows-1533.herokuapp.com (not fully up to date)


crowdsourcing/models.py
```diff
@@ -298,6 +298,12 @@ class Task(models.Model):
 class TaskWorker(models.Model):
     task = models.ForeignKey(Task, related_name='task_workers')
     worker = models.ForeignKey(Worker)
+    statuses = ((1, 'In Progress'),
```
dmorina: why are statuses needed here? basically the results have these statuses rather than the workers

aginzberg: because we want to accept a task rather than a taskworkerresult. it doesnt make sense to have to accept/reject each individual input item for a task (at least at first).

static/js/monitor/controllers/monitor.controller.js
```diff
@@ -137,11 +144,24 @@
       );
     }
 
-    function downloadResults(objects) {
-      var arr = [['status', 'last_updated', 'worker']];
-      for(var i = 0; i < objects.length; i++) {
-        var obj = objects[i];
-        var temp = [getStatusName(obj.status), obj.last_updated, obj.worker_alias];
+    function downloadResults(entries) {
```
dmorina: this should be generated in the backend

static/js/project/controllers/project.controller.js
```diff
@@ -58,8 +58,19 @@
       };
 
       self.myProjects = [];
+      self.myModules = [];
       Project.getRequesterProjects().then(function(data) {
         self.myProjects = data[0];
```
dmorina: this should be sufficient, looping in the templates will give you everything you need

static/js/project/controllers/project.controller.js
```diff
@@ -285,8 +296,8 @@
         return status == 1 ? 'created' : (status == 2 ? 'in review' : (status == 3 ? 'in progress' : 'completed'));
       }
 
-      function monitor(project) {
-        window.location = 'monitor/' + project.id;
+      function monitor(module) {
+        window.location = 'monitor/' + module.id +'?project=' + module.project + '&milestone=' + module.name;
```
dmorina: use angular $location if possible

crowdsourcing/models.py
dmorina: fair point, this would move to another model then, here we just want to keep track of who is working on the tasks

crowdsourcing/migrations/0026_taskworker_status.py
```diff
@@ -0,0 +1,19 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('crowdsourcing', '0025_auto_20150805_1718'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='taskworker',
+            name='status',
```
dmorina: let's name this task_status then we will be fine

crowdsourcing/serializers/project.py
```diff
@@ -66,7 +66,7 @@ def create(self, **kwargs):
                     'module': module.id,
                     'data': json.dumps(row)
                 }
-                task_serializer = TaskSerializer(data=task)
+                task_serializer = TaskSerializer(data=task, fields=('module', 'data'))
```
dmorina: do we need to do bottom up serialization, as in children serializing parent data?

crowdsourcing/serializers/project.py
```diff
@@ -76,7 +76,7 @@ def create(self, **kwargs):
                 'module': module.id,
                 'data': "{'type': 'static'}"
             }
-            task_serializer = TaskSerializer(data=task)
+            task_serializer = TaskSerializer(data=task, fields=('module', 'data'))
```
dmorina: same as above, not needed

crowdsourcing/viewsets/task.py
```diff
@@ -58,22 +58,18 @@ def create(self, request, *args, **kwargs):
             return Response(serializer.errors,
                             status=status.HTTP_400_BAD_REQUEST)
 
+    def update(self, request, *args, **kwargs):
```
dmorina: if this is a detail route then self.get_object() will do and update should happen in the serializer

static/js/monitor/services/monitor.service.js
```diff
@@ -37,21 +38,36 @@
...
     }
 
+    function getResultData(id, moduleName, projectName) {
+      var settings = {
+        url: 'api/csvmanager/get-results-file/',
+        method: 'GET',
+        params: {
+          id: id,
+          moduleName: moduleName,
+          projectName: projectName
```
dmorina: i dont think we need to send these around

dmorina:
will merge manually step by step

PR #360, opened 2015-08-13
Post prototype / adding future milestones

aginzberg:


static/js/project/services/project.service.js
```diff
@@ -56,7 +58,7 @@
           modules: [
```
neilthemathguy: wrt  ```name: 'Prototype Task'``` 
At present prototype task and future milestone are generic modules. Let's add a boolean flag to differentiate between both. This will help  related topic such as use of prototype task for ranking evaluation and so on. Backend model can be updated to hold this field.

dmorina:
let's leave category out for now, as it's a project property it's not loaded

crowdsourcing/viewsets/project.py
```diff
@@ -105,7 +105,14 @@ def destroy(self, request, *args, **kwargs):
 class ModuleViewSet(viewsets.ModelViewSet):
     queryset = Module.objects.all()
     serializer_class = ModuleSerializer
-    #permission_classes=[IsOwnerOrReadOnly, IsAuthenticated]
+    permission_classes=[IsOwnerOrReadOnly, IsAuthenticated]
+
+    @list_route(methods=['get'])
+    def get_prototype_task(self, request, **kwargs):
+        prototype_task = Module.objects.all().filter(project=self.request.query_params.get('projectId'), \
+                                                        name="Prototype Task")[0]
```
dmorina: we can take the first() milestone here instead of relying on the name, it's safe

PR #375, opened 2015-08-17
Reputation beginning

aginzberg:
migration and fixture data


crowdsourcing/models.py
```diff
@@ -502,10 +502,11 @@ def delete(self, *args, **kwargs):
 class WorkerRequesterRating(models.Model):
     origin = models.ForeignKey(UserProfile, related_name='rating_origin')
     target = models.ForeignKey(UserProfile, related_name='rating_target')
+    module = models.ForeignKey(Module, related_name='rating_module')
     weight = ((1, "BelowExpectations"),
               (2, 'MetExpectations'),
               (3, 'ExceedsExpectations'))
-    weight = models.IntegerField(choices=weight,default=2)
+    weight = models.FloatField(choices=weight,default=2)
```
dmorina: if we want other values that choices what do we do?  :P

PR #378, opened 2015-08-18
basic ranking for bipartite network 

aginzberg:
ranks feed purely by weight of projects' requesters by that particular worker.


dmorina:
for now only projects with ratings will show up, but ok, good start ;)

PR #381, opened 2015-08-19
Made atomic transaction in TaskWorkerSerializer.create() thread safe.

suvansh:
Added locking and unlocking inside TaskWorkerSerializer.create()

### DIFF
--- a/crowdsourcing/serializers/task.py
+++ b/crowdsourcing/serializers/task.py
@@ -33,6 +33,9 @@ class TaskWorkerResultSerializer(DynamicFieldsModelSerializer):
 
 
 class TaskWorkerSerializer(DynamicFieldsModelSerializer):
+    import threading
+    lock = threading.Lock()
+    
     task_worker_results = TaskWorkerResultSerializer(many=True, read_only=True)
     worker_alias = serializers.SerializerMethodField()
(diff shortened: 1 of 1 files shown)


crowdsourcing/serializers/task.py
```diff
@@ -48,23 +51,29 @@ def create(self, **kwargs):
...
-        with transaction.atomic(): # select_for_update(nowait=False)
-            tasks = models.Task.objects.filter(module=module).exclude(
-                task_workers__worker=kwargs['worker']) \
-                .annotate(task_worker_count=Count('task_workers')) \
-                .filter(module__repetition__gt=F('task_worker_count')).first()
-            if not tasks:
-                tasks = models.Task.objects.filter(module=module) \
-                    .exclude(task_workers__worker=kwargs['worker'], task_workers__task_status=6) \
+        
+        self.lock.acquire()
```
dmorina: does this work in multiprocessing?

dmorina:
great work, would be good to have a db level lock, or one across different processes because most likely we will serve this using multiple workers
in any case make sure you release() lock, Exceptions may be raised by anything in that region

dmorina:
merged with multiprocessing

PR #383, opened 2015-08-19
removed projects that worker created as requester

aginzberg:


dmorina:
Careful there, it should be requester id rather than user id, check model to confirm

dmorina:
looks good

PR #390, opened 2015-08-21
Authentication fix

dmorina:
@nistala we are rolling back to Session Auth until we have the middleware to encrypt oauth cookies, oauth2 is supported as well

@aginzberg can you test?


Macbull:
@dmorina : just to confirm,just before this PR, request.user was always Anonymous . Right ??

dmorina:
Hey, only if you weren't logged in, if yes then it wasn't Anonymous

mbernst:
I'm out of date. Was this addressed with the recent logged-in check PR?

dmorina:
@mbernst no, this is a different question, when unauthenticated the backend will return Anonymous when trying to access `request.user`, this is a non-issue

PR #397, opened 2015-08-24
Rating backend and UI

nistala:
Rating backend and UI

#447


dmorina:
checking..

static/templates/ranking/ranking.html
```diff
@@ -0,0 +1,87 @@
...
+            <div ng-repeat="item in ranking.requesterRankings">
+
+                <div>
+                    <div class="md-data-table-row" layout="row" layout-align="start center"
+                         >
+                        <div class="md-data-table-row-cell cell-text" flex="25">{{item.project.owner.alias}}</div>
+                        <div class="md-data-table-row-cell cell-text" flex="25">{{item.project.name}}</div>
+                        <div class="md-data-table-row-cell cell-text" flex="25">{{item.module_data.name}}</div>
+                        <div class="md-data-table-row-cell cell-text" flex>
+                            <div checkrating click-handler="ranking.handleRatingSubmit" model="item" selected="item.current_rating"></div>
```
dmorina: screen shot 2015-08-24 at 11 51 11

dmorina: can you move actions to the top rather than per row (let's follow the specs), thanks

crowdsourcing/viewsets/project.py
```diff
@@ -141,6 +142,81 @@ def requester_projects(self, request, **kwargs):
...
 
+    @list_route(methods=['GET'])
+    def workers_reviews(self, request, **kwargs):
+        projects = request.user.userprofile.requester.project_owner.all()
+        modules = Module.objects.all().filter(project__in=projects)
+        tasks = []
+        module_task_map = {}
+        for module in modules:
+          module_tasks = Task.objects.all().filter(module=module)
+          for tsk in module_tasks:
```
dmorina: no need to loop, use joins

crowdsourcing/viewsets/project.py
```diff
@@ -141,6 +142,81 @@ def requester_projects(self, request, **kwargs):
...
+          origin=request.user.userprofile, module__in=modules, type="requester")
+        rating_map = {}
+        for rating in ratings:
+          rating_map[(rating.module.id, rating.target.id)] = rating
+
+        for key, val in rating_map.items():
+          if key in pending_reviews:
+            current_review = pending_reviews[key]
+            current_review["current_rating"] = val.weight
+            current_review["current_rating_id"] = val.id
```
dmorina: multiple db hits and loops will kill perfomance;
use something like the following which achieves what you need:
```
from django.db.models import F, Count
TaskWorker.objects.values('task__module__name', 'task__module__id', 
                          'worker__profile__rating_target__weight', 'worker__profile__id')\
    .filter(task__module__project__owner=request.user.userprofile.requester.id, task_status__in=[3,4,5]).extra(
    where=["(crowdsourcing_module.id=crowdsourcing_workerrequesterrating.module_id OR crowdsourcing_workerrequesterrating.module_id is null)"]
).annotate(task_count=Count('task'))
```

crowdsourcing/viewsets/project.py
```diff
@@ -141,6 +142,81 @@ def requester_projects(self, request, **kwargs):
...
+
+        return Response(pending_reviews.values())
+
+
+    @list_route(methods=['GET'])
```
dmorina: should be in your rating viewset, or worker/requester viewset respectively

crowdsourcing/viewsets/project.py
```diff
@@ -141,6 +142,81 @@ def requester_projects(self, request, **kwargs):
         serializer = ProjectSerializer(instance=projects, many=True, fields=('id', 'name', 'module_count'))
         return Response(serializer.data)
 
+    @list_route(methods=['GET'])
```
dmorina: move to either rating or worker viewset

crowdsourcing/viewsets/project.py
```diff
@@ -141,6 +142,81 @@ def requester_projects(self, request, **kwargs):
...
+        ratings = WorkerRequesterRating.objects.all().filter(
+          origin=worker.profile, module__in=modules, type="worker")
+        rating_map = {}
+        for rating in ratings:
+          rating_map[(rating.module.id, rating.target.id)] = rating
+        for key, val in rating_map.items():
+          if key in pending_reviews:
+            current_review = pending_reviews[key]
+            current_review["current_rating"] = val.weight
+            current_review["current_rating_id"] = val.id
```
dmorina: same as above

static/js/ranking/controllers/ranking.controller.js
```diff
@@ -6,36 +6,75 @@
     .module('crowdsource.ranking.controllers')
     .controller('RankingController', RankingController);
 
-  RankingController.$inject = ['$scope', '$log', '$mdToast', '$http', 'RankingService'];
```
neilthemathguy: I think we should keep the turkopticon rating example. It has helped new folks getting started with our framework. (may be rename it to something else for now)

neilthemathguy:
**Requester to Worker**
worker-requester

requester to worker

crowdsourcing/viewsets/rating.py
```diff
@@ -0,0 +1,113 @@
...
+        rating_map = {}
+        for rating in ratings:
+          rating_map[(rating.module.id, rating.target.id)] = rating
+
+        for key, val in rating_map.items():
+          if key in pending_reviews:
+            current_review = pending_reviews[key]
+            current_review["current_rating"] = val.weight
+            current_review["current_rating_id"] = val.id
+
```
dmorina: see the outdated diff comment on this

maniksingh92:
#447

nistala:
@neilthemathguy added back turkopticon page
@dmorina updated

crowdsourcing/serializers/rating.py
```diff
@@ -0,0 +1,16 @@
...
+class WorkerRequesterRatingSerializer(serializers.ModelSerializer):
+    origin = UserProfileSerializer(read_only=True)
+
+    class Meta:
+        model = models.WorkerRequesterRating
+        fields = ('id', 'origin', 'target', 'module', 'weight',
+                  'type', 'created_timestamp', 'last_updated')
+
+    def create(self, **kwargs):
+        print kwargs
```
dmorina: rm print

crowdsourcing/serializers/rating.py
```diff
@@ -0,0 +1,16 @@
...
+from crowdsourcing.serializers.user import UserProfileSerializer
+from rest_framework import serializers
+
+class WorkerRequesterRatingSerializer(serializers.ModelSerializer):
+    origin = UserProfileSerializer(read_only=True)
+
+    class Meta:
+        model = models.WorkerRequesterRating
+        fields = ('id', 'origin', 'target', 'module', 'weight',
+                  'type', 'created_timestamp', 'last_updated')
```
dmorina: origin, 'created_timestamp', 'last_updated' are read_only

crowdsourcing/viewsets/rating.py
```diff
@@ -0,0 +1,119 @@
...
+            "worker_profile_id": review["worker__profile__id"]
+          }
+          response.append(row)
+
+        return Response(response)
+
+
+    @list_route(methods=['GET'])
+    def requesters_reviews(self, request, **kwargs):
+        worker = Worker.objects.get(profile=request.user.userprofile)
```
dmorina: fyi: here you can use request.user.userprofile.worker

static/js/dashboard/controllers/dashboard.controller.js
```diff
@@ -114,6 +120,68 @@
               $mdToast.showSimple('Drop tasks failed.')
             }
         ).finally(function () {});
+      }
+
+    function getWorkerData() {
```
dmorina: let's remove this from Dashboard

static/js/dashboard/controllers/dashboard.controller.js
```diff
@@ -114,6 +120,68 @@
...
+          self.requesterRankings = data;
+        },
+        function error (errResp) {
+          var data = resp[0];
+          $mdToast.showSimple('Could not get requester rankings.');
+        });
+    }
+
+    function refreshData(reviewType) {
+      if (reviewType === 'worker') {
```
dmorina: this should only handle worker rating requesters

static/js/ranking/services/ranking.service.js
```diff
@@ -29,6 +33,56 @@
...
+      };
+      return HttpService.doRequest(settings);
+    }
+
+    function updateRating(rating, entry) {
+      var settings = {
+        url: '/api/worker-requester-rating/' + entry.current_rating_id + '/',
+        method: 'PUT',
+        data: {
+          weight: rating,
```
dmorina: since you are using current_rating_id only weight is required here, you may set partial=True on update on backend

nistala:
@dmorina backend adressed
let's keep worker ratings stuff on the frontend for reference

dmorina:
screen shot 2015-08-31 at 21 28 53

dmorina:
if i keep clicking the rate requester button it just keeps creating ratings for the same module same requester

dmorina:
add the sentences to on top of the rating: below Requester Rating subheader
the famous two from the wiki :)

nistala:
@dmorina addressed

dmorina:
could you replace the updated data into the object rather than fetching everything on every submit/update?

nistala:
@dmorina  let's not delay this PR for optimizations, let me know if there is broken functionality

dmorina:
@nistala let's treat all PRs the same way, we can take the time to fix all addressed issues we have tomorrow as well, if 461 had a problem with a commented out line and an unused variable, or the previous ones not showing a message which is a UX problem, we should do the same here

dmorina:
type must be 'worker' not requester

dmorina:
please see my last comment `type must be 'worker' not requester`  the type shows from whom the rating is coming

nistala:
it's unclear in the model that type refers to who it is coming from, how can we make this clearer?

dmorina:
we can rename to `origin_type`

nistala:
sure, what would break if we left it as is?

dmorina:
so, the task feed wouldnt be sorted correctly if type is wrong

nistala:
I can fix but this is a large change, i highly recommend naming models correctly going forward

nistala:
updated

dmorina:
breaks task feed as all the queries rely on type, let's not rename it for now or we fix all other queries which is way more work

dmorina:
it shouldnt be a problem if you just submit `worker` instead of `requester`

mbernst:
Hey all, I'm setting a goal of being a bit more involved in PRs at least to observe. Let me know if I should back off.

Sounds like the main question here is whether it's worth fixing the architecture of the feed now, or waiting until after CHI. All research systems have hacky parts to them, you just want to make sure you’re making those technical debt tradeoffs intelligently. Basically, will this bite us significantly as we attempt to build the last few features or debug them in the deployment?

So @dmorina, how long do you think it would take a pro like you to fix the underlying issue with the queries in the feed, in terms of hours? And would we run up against the problem again soon?

Tagging @neilthemathguy and @aginzberg to weigh in on the decision.

dmorina:
@mbernst there is a field name which is not self explanatory that's all
renaming it for the feed should take 2-4mins incl. testing, but anyways the right value needs to be submitted..

dmorina:
@mbernst all set, see #466, it shouldn't happen again, I think this is the only field name that could have confused people, im happy to go through them again

mbernst:
thanks for looking into this @dmorina

static/js/ranking/services/ranking.service.js
dmorina: Here you only submit weight..

dmorina:
Origin is read only

nistala:
readonly issue addressed 9 days ago -https://github.com/nistala/crowdsource-platform/blob/develop2/crowdsourcing/serializers/rating.py#L6 @dmorina

dmorina:
Lgtm ( remaining comments to be addressed with patches after merge)

neilthemathguy:
Good to go - just add the migrations back

PR #403, opened 2015-08-25
backend to fetch task populated with results

aginzberg:


crowdsourcing/serializers/task.py
```diff
@@ -96,6 +97,19 @@ def get_module(self, obj):
...
+    def get_task_with_data_and_results(self, obj):
+        task = TaskSerializer(instance=obj.task, fields=('id', 'task_template')).data
+        template = task['task_template']
+        task_worker_results = TaskWorkerResultSerializer(instance=obj.task_worker_results, many=True,
+                                                            fields=('template_item_id', 'result')).data
+        for task_worker_result in task_worker_results:
+            for item in template['template_items']:
+                if task_worker_result['template_item_id'] == item['id'] and item['role'] == 'input' \
+                and task_worker_result['result'] is not None and item['type'] != 'labeled_input':
+                    item['values'] = task_worker_result['result']
```
dmorina: does this match to item.answer in the front end?

aginzberg: well not exactly, for example, we do not save for labeled input. pretty sure it matches for the relevant template items like text field and text area. maybe i should just specify those ones to replace?

dmorina: here you want to send back  field answer, so that front end keeps label = values and answer is the actual value

aginzberg: so dont replace the values and let front end do the work to populate the answer?

dmorina: values as before, answer needs to be added

dmorina:
lgtm

PR #407, opened 2015-08-25
Worker Dashboard

aginzberg:
Midway with worker dashboard. TODO:
1. actually populate data table with the values we get for in progress and accepted tasks
2. allow user to check out in progress tasks and complete or skip them (go back to accept task or stay on dashboard?)
3. log and display payment information
4. Post microsoft - allow workers to rate requesters for their accepted tasks


dmorina:
lgtm, please include returned tasks as well :)

PR #415, opened 2015-08-25
Tablefrontend

aginzberg:
This is the start of the in progress table front end. Left todo is when user clicks droppedsaved task / get saved task icon update the backend and table.


dmorina:
lgtm

PR #416, opened 2015-08-25
fix no tasks left front end

dmorina:


dmorina:
@aginzberg  Adaaaaaaaam :D

aginzberg:
lgtm but still a few things left.

PR #417, opened 2015-08-25
delete tasks from dashboard only

aginzberg:


crowdsourcing/viewsets/task.py
```diff
@@ -134,11 +134,13 @@ def retrieve_with_data_and_results(self, request, *args, **kwargs):
...
-    def drop_saved_task(self,request, *args, **kwargs):
-        obj = self.queryset.get(task=kwargs['task__id'], worker=request.user.userprofile.worker.id)
-        obj.task_status = 6
-        obj.save()
+    @list_route(methods=['post'])
+    def drop_saved_tasks(self,request, *args, **kwargs):
+        task_ids = request.data.get('task_ids', [])
+        for task_id in task_ids:
+            obj = self.queryset.get(task=task_id, worker=request.user.userprofile.worker.id)
+            obj.task_status = 6
```
dmorina: do a bulk_update here no need to loop

self.queryset.filter(id__in=task_ids).update(....)

dmorina: see bulk_update_status method here

PR #419, opened 2015-08-26
Saved and deleted in progress tasks on dashboard

aginzberg:


crowdsourcing/viewsets/task.py
```diff
@@ -131,14 +131,13 @@ def list_by_status(self, request, *args, **kwargs):
...
 
-    @detail_route(methods=['delete'])
-    def drop_saved_task(self,request, *args, **kwargs):
-        obj = self.queryset.get(task=kwargs['task__id'], worker=request.user.userprofile.worker.id)
-        obj.task_status = 6
-        obj.save()
+    @list_route(methods=['post'])
+    def drop_saved_tasks(self,request, *args, **kwargs):
+        task_ids = request.data.get('task_ids', [])
+        self.queryset.filter(task_id__in=task_ids, worker=request.user.userprofile.worker.id).update(task_status=6)
```
dmorina: set last_udated=timezone.now bitte

dmorina:
lgtm for now, please add status to the table so that worker knows that sth is changing
repeating code to be optimized

crowdsourcing/viewsets/task.py
aginzberg: danke

aginzberg:
do we want status in the table? all of the tasks in this table are in progress as it notes on the header. after the worker takes an action on the task it will either disappear from the table or not change status

dmorina:
alright, then just a notification would help 'Changes saved' or sth

PR #424, opened 2015-08-26
fixed tasks available query

dmorina:


crowdsourcing/serializers/project.py
```diff
@@ -105,30 +105,39 @@ def get_total_tasks(self, obj):
...
 
     def get_has_comments(self, obj):
-        return obj.modulecomment_module.count()>0
+        return obj.modulecomment_module.count() > 0
 
     def get_available_tasks(self, obj):
-        available_task_count = obj.module_tasks.exclude(Q(task_workers__worker=self.context['request'].user.userprofile.worker)
-                                                        |Q(task_workers__task_status__in=[4, 6])).annotate(task_worker_count=Count('task_workers'))\
-            .filter(module__repetition__gt=F('task_worker_count')).count()
+        available_task_count = models.Module.objects.values('id').raw('''
```
dmorina: this has changed only

nistala: what's changed here?

crowdsourcing/serializers/project.py
```diff
@@ -105,30 +105,39 @@ def get_total_tasks(self, obj):
...
+          select count(*) id from (
+            SELECT
+              "crowdsourcing_task"."id"
+            FROM "crowdsourcing_task"
+              INNER JOIN "crowdsourcing_module" ON ("crowdsourcing_task"."module_id" = "crowdsourcing_module"."id")
+              LEFT OUTER JOIN "crowdsourcing_taskworker" ON ("crowdsourcing_task"."id" = "crowdsourcing_taskworker"."task_id" and task_status not in (4,6))
+            WHERE ("crowdsourcing_task"."module_id" = %s AND NOT (("crowdsourcing_task"."id" IN (SELECT U1."task_id" AS Col1
+                                                                                                FROM "crowdsourcing_taskworker" U1
+                                                                                                WHERE U1."worker_id" = %s))))
```
aginzberg: I think i tracked it down. WHERE U1."worker_id"=%s AND task_status not in (4,6). We want to exclude tasks where the worker is the worker on a task worker for the task but only when it is not returned or skipped.

dmorina: Why?
So worker would see the tasks he did as available?

dmorina: Got it

PR #431, opened 2015-08-26
allow user to save partial results

aginzberg:


crowdsourcing/serializers/task.py
```diff
@@ -18,6 +18,7 @@ def create(self, **kwargs):
 
 class TaskWorkerResultSerializer(DynamicFieldsModelSerializer):
     template_item_id = serializers.SerializerMethodField()
+    result = serializers.CharField(required=False, allow_blank=True)
```
dmorina: does the model allow blanks?

aginzberg: it does as a matter of fact

dmorina: good to go then ;)

PR #432, opened 2015-08-27
Change password

dmorina:
front end for change password


aginzberg:
lgtm

note: when you type in old password incorrectly the validation message says "username or password is incorrect" but i think it should just say "old password is incorrect"

dmorina:
got it

PR #446, opened 2015-08-28
Microsoft bug fixes

aginzberg:
Skip button on one task available -> returns to task feed
Submit only completed task (can still save partial)
Check all on dashboard (still need to allow worker to return to multiple)
Fix no more tasks available so it doesnt flicker and is displayed when appropriate

### DIFF
--- a/static/templates/task-feed/main.html
+++ b/static/templates/task-feed/main.html
@@ -72,11 +72,13 @@
                 </div>
             </md-card>
         </div>
-        <md-card ng-if="!taskfeed.projects.length">
-            <md-card-content>
-                <h2 class="md-title">No free tasks available</h2>
-            </md-card-content>
-        </md-card>
+        <div ng-show="taskfeed.availableTasks">
+            <md-card ng-if="!taskfeed.availableTasks">
+                <md-card-content>
+                    <h2 class="md-title">No free tasks available</h2>
+                </md-card-content>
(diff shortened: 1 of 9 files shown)


neilthemathguy:
lgtm++; we should still think through the flow here. 
@dmorina, @nistala, @ksetyadi take a look. Two more lgtm and we can merge the PR.

dmorina:
Lgtm

nistala:
lgtm

nistala:
#446

static/templates/task-feed/main.html
```diff
@@ -72,11 +72,13 @@ <h3 class="comment-sender">{{ item.comment.sender_alias }}</h3>
                 </div>
             </md-card>
         </div>
-        <md-card ng-if="!taskfeed.projects.length">
-            <md-card-content>
-                <h2 class="md-title">No free tasks available</h2>
-            </md-card-content>
-        </md-card>
+        <div ng-show="taskfeed.availableTasks">
```
dmorina: remove one of these ng-shows, @maniksingh92 pointed this out

aginzberg: @maniksingh92, did you test this? I actually did this intentionally to get rid of the annoying flicker. So 'no more tasks left' only shows up once we we're sure there are none left.

Issue #443, opened 2015-08-28
Registration workflow 


neilthemathguy:
Change the registration flow:  (reported by Jsilver) 
* User will signup using email 
* We send confirmation link 
* User sets the password

User story 
> I registered using a fake/nonexistent email. The system pushed through with my registration

Setting this as low priority for now. Registration will be controlled for next few weeks. Follow #409

PR #449, opened 2015-08-29
Registration activation link

dmorina:
addresses #443
activation links are sent via email, as long as we are on Dev/Staging we can use Gmail, for production we switch to SendGrid


aginzberg:
works fine for me. activation key does the trick. couldnt send email at first because gmail was intercepting because unknown account (we need to fix this). also when production comes around we cant keep login credentials in settings.

dmorina:
For production we will use SendGrid instead

PR #456, opened 2015-08-30
Forgot password

dmorina:
adds forgot password functionality
new button on login, sends email with reset_key or ignore link


aditimithal:
I was trying to sign-in but it shows "Account is not activated "
On clicking the forgot password button it sends a link and and i am able to change the password too.But ,i am unable to login and it shows that the account is not activated. How do i go about it?
Also, will there be a message saying "invalid email address" if i try to add an invalid email-id ?

dmorina:
have you received an activation link initially when you singed up?

aditimithal:
I was an existing user so i did not register again :P So,
Entered Valid email-> I registered again , got the activation link and everything is working well :D
Invalid email-> I could not find a message that shows "invalid email". I hope the email i entered does not exist , because i chose a really weird name.(gmail account)

aditimithal:
The email i chose does not exist(the test message bounced).

dmorina:
ok, so the process is as follows you register, then you will get an activation link you follow it and it will activate your account, then you can use forgot password with the same email and you follow the other link :) it should work

aditimithal:
That works if you have entered a valid email address :+1:

nistala:
After I reques

nistala:
clicked wrong button while commenting.

static/templates/authentication/forgot-password.html
```diff
@@ -0,0 +1,17 @@
+<div  layout="row" flex="50" layout-align="center center">
+    <md-content class="authentication-form" layout-padding class="md-gray-theme" flex="50" layout-align="center center" layout="column">
+        <h3 id="form-header">Forgot Password</h3>
+        <form name="form">
+            <div layout="row">
+                <md-input-container>
+                    <label>Email</label>
+                    <input type="email" required ng-model="auth.email">
+                </md-input-container>
```
nistala: after I submit the email address, i am on the same page, we should show something like "email sent" in a toast message or on the form.

nistala:
Go to forgot password page
hit submit twice
click on first link and click enter new email and password
hit enter and I don't see an error message - ideally should say something like "looks like an expired request"

static/js/authentication/controllers/auth-settings.controller.js
```diff
@@ -18,11 +18,18 @@
         var self = this;
 
         self.changePassword = changePassword;
+        self.resetPassword = resetPassword;
+        self.submitForgotPassword = submitForgotPassword;
 
         activate();
         function activate() {
             if (!Authentication.isAuthenticated() && $location.path().match(/change-password/gi)) {
```
nistala: can we add a link for change-password on my account page

dmorina: we will add it in the dropdown next to the username on top right corner

maniksingh92:
lgtm functionality wise.

dmorina:
@nistala fixed success message and the other issue with invalid reset_key

nistala:
lgtm

neilthemathguy:
lgtm

PR #463, opened 2015-08-31
Repeat rating fix

aginzberg:
This addresses if multiple rating pairs (same worker same requester) exist in the workerrequesterrating table. This will potentially occur often and is important to resolve. Fixes for cascading and taskfeed. The contents of the first LEFT OUTER JOIN removes all old ratings. The contents of the second LEFT OUTER JOIN corrects the average rating calculation so that we do not include our own ratings in the average and so that it is computed only using most recent ratings.  

Please check carefully. Very easy to get bogged down in sql. Comments much appreciated.


crowdsourcing/viewsets/project.py
```diff
@@ -89,11 +89,25 @@ def list(self, request, *args, **kwargs):
...
+                    SELECT origin_id, MAX(last_updated) AS max_date FROM crowdsourcing_workerrequesterrating
+                        WHERE type='requester' AND target_id = %s GROUP BY origin_id) tb
+                    ON w.origin_id = tb.origin_id AND w.last_updated = tb.max_date
+                    AND w.type='requester' AND w.target_id=%s) w
+                ON u.id = w.origin_id
+            LEFT OUTER JOIN (
+                SELECT target_id, AVG(weight) AS average_worker_rating from
+                (SELECT wr.* FROM crowdsourcing_workerrequesterrating wr
+                INNER JOIN (
+                    SELECT origin_id, target_id, MAX(last_updated) AS max_date
```
dmorina: you wanna exclude `this` requester from the calculated average I suppose?

crowdsourcing/viewsets/project.py
```diff
@@ -117,18 +131,27 @@ def list(self, request, *args, **kwargs):
...
+    ON up.id = wrr.target_id
+LEFT OUTER JOIN (
+    SELECT target_id, AVG(weight) AS average_requester_rating from
+    (SELECT wr.* FROM crowdsourcing_workerrequesterrating wr
+    INNER JOIN (
+        SELECT origin_id, target_id, MAX(last_updated) AS max_date FROM crowdsourcing_workerrequesterrating
+        GROUP BY origin_id, target_id) fltr
+    ON fltr.origin_id=wr.origin_id AND fltr.target_id=wr.target_id AND wr.last_updated=fltr.max_date
+    AND wr.origin_id <> %s AND wr.type='worker') sult GROUP BY target_id) avg
+ON avg.target_id = up.id) calc WHERE owner_id<>%s
```
dmorina: this part lgtm

crowdsourcing/viewsets/project.py
aginzberg: so i think that makes a little bit less sense in the cascade than in the task feed. but yes i suppose we could exclude this user acting as a requester rating himself as a worker. 

thinking it through more, it is impossible because we dont allow users (as workers) to accept tasks they posted as requesters and therefore garner ratings from themselves. so shouldnt be a problem? what do you think?

mbernst: All things equal, best to exclude this worker in case we forget to enforce
that constraint later? Unless it's a big performance hit.

(Again, trying to be a bit more present in the PR system here, lemme know
if I'm out of line)

dmorina: fyi `this` requester means project_owner

dmorina: @mbernst yep, so we calculate the worker's rating from the project owner and the second part should calculate the average by excluding the project_owner

crowdsourcing/viewsets/project.py
```diff
@@ -67,68 +67,103 @@ def list(self, request, *args, **kwargs):
...
+                        wr.last_updated=fltr.max_date AND wr.target_id=%s AND wr.origin_type='requester') sult
+                    GROUP BY target_id) avgreq
+                ON w.target_id=avgreq.target_id
+                INNER JOIN (
+                SELECT origin_id, target_id, MAX(last_updated) AS max_date
+                        FROM crowdsourcing_workerrequesterrating
+                    GROUP BY origin_id, target_id
+                ) tmp ON w.origin_id = tmp.origin_id AND w.target_id = tmp.target_id AND
+                        w.last_updated=tmp.max_date AND w.origin_type='requester') temp) als
+                ON owner_id=als.origin_id
```
dmorina: at some point you have to filter by 
owner_id <> origin_id when calculating the average rating of the worker

aginzberg: Note that the average is calculated using all weights and then adjusted for each particular origin afterwards. so the adjusted rating is unique to each origin/target pair. where the target specified is always the worker in question.

crowdsourcing/viewsets/project.py
```diff
@@ -67,68 +67,103 @@ def list(self, request, *args, **kwargs):
...
+                INNER JOIN(
+                    SELECT origin_id, MAX(last_updated) AS max_date FROM crowdsourcing_workerrequesterrating
+                        WHERE origin_type='requester' AND target_id = %s GROUP BY origin_id) tb
+                    ON w.origin_id = tb.origin_id AND w.last_updated = tb.max_date
+                    AND w.origin_type='requester' AND w.target_id=%s) w
+                ON u.id = w.origin_id
+            LEFT OUTER JOIN (
+                SELECT temp.origin_id, temp.target_id, temp.average_worker_rating, temp.count, temp.weight,
+                     (temp.average_worker_rating * temp.count - temp.weight) /
+                     (temp.count-1) as adj_average_worker_rating FROM
```
dmorina: gotcha,  cool
simplification suggestion no need to be considered now just in case you are interested
you can do a left outer join to wrrating on <your existing predicates> and origin_id<> owner_id, similarly as you did below, reduces number of rows processed and decreases the cost

dmorina:
lgtm

neilthemathguy:
lgtm.

PR #465, opened 2015-09-01
User override

aginzberg:
This allows a user logged in as an admin to spontaneously create more users (while specifying worker or requester or both). To test you can make yourself an admin by UPDATE auth_user SET is_staff='t' WHERE id={your_id};


crowdsourcing/serializers/user.py
```diff
@@ -135,15 +137,18 @@ def create(self, **kwargs):
...
-        worker.profile = user_profile
-        worker.alias = username
-        worker.save()
-
-        requester = models.Requester()
-        requester.profile = user_profile
-        requester.alias = username
-        requester.save()
+
+        if 'is_worker' not in self.validated_data or self.validated_data['is_worker']:
```
dmorina: im assuming you meant `in`

crowdsourcing/serializers/user.py
```diff
@@ -135,15 +137,18 @@ def create(self, **kwargs):
...
+            worker = models.Worker()
+            worker.profile = user_profile
+            worker.alias = username
+            worker.save()
+
+        if 'is_requester' not in self.validated_data or self.validated_data['is_requester']:
```
dmorina: same as above

dmorina:
there is a validator crowdsourcing/validators/utils.py RegistrationAllowed
which will invalidate this action, you might have to change that

crowdsourcing/serializers/user.py
aginzberg: not quite. basically if these fields arent specified (i.e. via normal user creation) then 'is_worker' not in self.validated_data is True and  the Worker should be created by default. Otherwise self.validated_data['is_worker'] will exist and the Worker object will be created appropriately. Kind of a short circuit dealio.

aginzberg:
thanks ill look into it later tonight.

crowdsourcing/serializers/user.py
dmorina: gotcha, so by default neither will be created, it's a part of profile setup later (this was temporary for UIST)

aginzberg: no actually right now the user is created as a worker and requester by default (as before). is that what we want? if we dont, then you are right it should be **in** not **not in**.

dmorina: Yep, creating profiles was not really supposed to be here, by default a user is neither I know now we are making him both ;)

neilthemathguy:
Let's complete registration close as well.

dmorina:
lgtm

neilthemathguy:
lgtm.

mbernst:
thank you!

PR #467, opened 2015-09-01
Task Comments, SCSS merged

shirishgoyal:
image


dmorina:
can you make the *comments* area overflow: scroll and set a max height too 100% of the page 
otherwise lgtm

dmorina:
You can ignore the comment about overflow ;) 
Lgtm

neilthemathguy:
This is great and covers the workers side of the feedback capture. However, we will need to aggregate and communicate the comments to the requester so that he can improve the task. This will help us measure the effect of feedback.This can go under 1) the MyProject - in front of the tasks 2)  In the result cvs file 3) on the Design dashboard of next Milestone as third tab in the control panel: 

Library Properties Feedback 
feedback


Also _Task Comment_ doesn't communicate much; we can change it to _Please provide feedback on the task design_ or something along the line

shirishgoyal:
Ok

When you say next milestone, is it referring all milestones after prototype one or only immediate one ?

neilthemathguy:
Thats correct!

shirishgoyal:
How do we want to club comments within CSV? 
Comma seperated comments per task in row
OR add comments underneath task results (column consistency will be lost in this case though)

neilthemathguy:
For MSFT and CHI we can go with per task in a row.

shirishgoyal:
image

dmorina:
@shirishgoyal what types of comments are being shown below the table, all task comments? Here it would make more sense if it's Module comments or one level up actually

shirishgoyal:
image

shirishgoyal:
image

dmorina:
screen shot 2015-09-02 at 00 33 13
screen shot 2015-09-02 at 00 33 20

dmorina:
@shirishgoyal there is a problem with loading and rendering comments, see above

nistala:
very long comment lines squished together, steps to reproduce:
Go to comment page
Enter long comment (more than 3 lines)
submit
screenshot from 2015-09-01 18 26 38

shirishgoyal:
1st issue happened due to merge of other PRs.
Resolved both now.

image

image

neilthemathguy:
lgtm.

static/templates/base/index.html
```diff
@@ -39,26 +32,38 @@ <h1 layout-align="left"><a href="/">DAEMO</a></h1>
     </md-toolbar>
 </div>
 
-<div ng-if="!isLoggedIn" id="topcontainer" style="background-color: #E91E63;"><canvas id="topanvas" width="1417" height="8"></canvas></div>
-<div ng-if="!isLoggedIn" id="secondcontainer" style="background-color: #000000;"><canvas id="secondcanvas" width="1417" height="5"></canvas></div>
+<div ng-if="!isLoggedIn" id="sub" ng-cloak>
+    <div id="topcontainer" style="background-color: #E91E63;">
```
nistala: I realize this PR did not add the inline styles but since we're working on CSS, could we remove these inline styles, most should not be necessary.

static/templates/template/container.html
```diff
@@ -1,8 +1,8 @@
-<md-subheader class="md-no-sticky">Use our Task Builder to design your task. Arrange and customize your task using our draggable template items. Double click a component to modify its properties.</md-subheader>
+<md-subheader class="md-no-sticky" ng-init="project.getLastMilestoneComments()">Use our Task Builder to design your task. Arrange and customize your task using our draggable template items. Double click a component to modify its properties.</md-subheader>
```
nistala: you can call this in the controller.

static/templates/base/index.html
shirishgoyal: I actually wanted to remove all this section as it is done in a hackish way. I will do that in next push.

static/templates/template/container.html
shirishgoyal: We do not want to keep calling this API every time tab changes, so it was called using ng-init only when tab - Design is loaded.

mbernst: Is this because there is no controller for the design tab?

shirishgoyal:
This is because tab content area is being manually maintained based on tab selected rather than using material design (MD) tab, so they all are part of larger scope.

static/templates/template/container.html
nistala: @shirishgoyal can we load it once when we load the controller, i can't think of too many instances when a reviewer won't want to see the comments.

nistala: @mbernst this makes it so that we don't load comments unless a person clicks on the tab. 
@shirish we can show the number of comments in the heading if we load it in the controller, I think that will be more helpful to the user.

dmorina:
Lgtm

shirishgoyal:
image

mbernst:
We have lgtm's from neil and durim, can we merge?

dmorina:
conflicts, can you resolve @shirishgoyal then we can merge right away

dmorina:
merged via command line ;) no action required

PR #469, opened 2015-09-01
Task Feed Flicker issue

shirishgoyal:
Issue by @jsilver
https://lit-hollows-1533.herokuapp.com/task-feed shows "No free tasks available" for a second 

Resolved

### DIFF
--- a/static/js/task-feed/controllers/task-feed.controller.js
+++ b/static/js/task-feed/controllers/task-feed.controller.js
@@ -32,7 +32,7 @@
       self.openTask = openTask;
       self.openComments = openComments;
       self.saveComment = saveComment;
-      self.availableTasks = true;
+      self.loading = true;
 
       TaskFeed.getProjects().then(
         function success (data) {
@@ -52,7 +52,7 @@
           $mdToast.showSimple('Could not get task with data.');
         }
       ).finally(function () {
-        // pass
+        self.loading = false;
       });
 
       function toggleBookmark(project){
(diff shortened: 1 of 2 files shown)


mbernst:
Interesting. Is this the intended behavior of ng-show? Wonder why they needed both.

shirishgoyal:
ng-if is preferred where we don't want to add elements to DOM, which is true in this case as once we know tasks exist, we no longer need that section in DOM at all.

But magic here is in JS file than ng-if/ng-show

dmorina:
Lgtm, ng-show is preferred when you want to hide show based on user actions/data so transitions would be ok, but for data loading ng-if is what we need with correct conditions 😛

nistala:
we should be showing a loading spinner here or some kind of message then hide it to show the tasks, transitions are bit overkill but this is where we would use ngAnimate.

shirishgoyal:
Loading spinner added

neilthemathguy:
lgtm. Thanx!

mbernst:
Thanks!

On Tuesday, September 1, 2015, neilthemathguy <notifications@github.com>
wrote:

> Merged #469
> <https://github.com/crowdresearch/crowdsource-platform/pull/469>.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/crowdresearch/crowdsource-platform/pull/469#event-398397939>
> .
>


-- 
sent from my NLS

Issue #479, opened 2015-09-02
Issues with Task Feed on Dashboard


shirishgoyal:
a) When there is no Task in Progress, once we select checkbox for all, we cannot deselect it

b) On selecting any task in Returned Tasks, it shows action buttons on Task in Progress section rather than Returned Tasks Section. See image below for reference.

![image](https://cloud.githubusercontent.com/assets/315012/9625955/a001a8ce-5177-11e5-957a-69c4e6c94efb.png)

dmorina:
A) has been addressed 
I agree b) should be like you described it

aginzberg:
regarding a. i think we should disable (or modify) the select all checkbox (like on github) when people start selecting individual tasks. we get some really weird behavior when you start selecting individual individual tasks and then start toggling the select all checkbox.

lucamatsumoto:
I found two "md-data-table" implementations and both work in the same way when you select check-all after selecting some individual rows. Demo sites; http://danielnagy.me/md-data-table/ http://iamisti.github.io/md-data-table/ We can implement the same behavior quickly if that is what we wanted. And we should do the same for the task overview.

shirishgoyal:
In progress

shirishgoyal:
Resolved a) which appears natural in function now,
and b)

shirishgoyal:
Resolved
b) Returned tasks maintain their separate queue. On save, task gets merged to task in progress list. 
Actions shown separately for Returned tasks which do not interfere with Task in progress queue.

PR #481, opened 2015-09-02
Deselection of checkbox

radhikabhanu:
Using ng-change over ng-click as ng-change always reflects the correct value of ng-model. With ng-click, the value of ng-model is ambiguous.


radhikabhanu:
Related issue: #479, part A.

dmorina:
lgtm

neilthemathguy:
lgtm.

lucamatsumoto:
The behavior has changed. When some of the checkboxes are checked, if you click the checkbox in the header, it was clearing all checkboxes including the header. But now, the checkbox in the header is still checked.

PR #488, opened 2015-09-03
rating optimizations

dmorina:
restricts object permissions only to the owner of the rating entry
task_worker removed from serialization as it's not needed
more optimizations needed for rating viewset


nistala:
lgtm

PR #491, opened 2015-09-05
Issues with Task Feed on Dashboard #479

shirishgoyal:
Resolved both a) and b)

a) When there is no Task in Progress, once we select checkbox for all, we cannot deselect it

b) On selecting any task in Returned Tasks, it shows action buttons on Task in Progress section rather than Returned Tasks Section. See image below for reference.

### DIFF
--- a/static/js/task/controllers/task.controller.js
+++ b/static/js/task/controllers/task.controller.js
@@ -22,17 +22,22 @@
             self.task_worker_id = $routeParams.taskWorkerId;
             self.task_id = $routeParams.taskId;
 
-            if (!self.task_worker_id) { //if they navigate away mid-queue and then attempt task from task feed
-                Dashboard.savedQueue = [];
-            } else if (Dashboard.savedQueue == undefined) { //if they refresh page mid-queue
+            self.isReturned = $routeParams.hasOwnProperty('returned');
+
+            if ((self.isReturned && Dashboard.savedReturnedQueue == undefined) || (!self.isReturned && Dashboard.savedQueue == undefined)) { //if they refresh page mid-queue
                 $location.path('/dashboard');
                 return;
             }
 
+            Dashboard.savedQueue = Dashboard.savedQueue || [];
+            Dashboard.savedReturnedQueue = Dashboard.savedReturnedQueue || [];
+
             self.isSavedQueue = !!Dashboard.savedQueue.length;
+            self.isSavedReturnedQueue = !!Dashboard.savedReturnedQueue.length;
+
             var id = self.task_worker_id ? self.task_worker_id : self.task_id;
 
-            Task.getTaskWithData(id, self.isSavedQueue).then(function success(data) {
+            Task.getTaskWithData(id, self.isSavedQueue || self.isSavedReturnedQueue).then(function success(data) {
                     self.taskData = data[0];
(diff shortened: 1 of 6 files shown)


static/js/task/controllers/task.controller.js
```diff
@@ -22,17 +22,22 @@
             self.task_worker_id = $routeParams.taskWorkerId;
             self.task_id = $routeParams.taskId;
 
-            if (!self.task_worker_id) { //if they navigate away mid-queue and then attempt task from task feed
```
aginzberg: see deleted comment. this now breaks if you stop mid-queue and then go back to task feed and attempt the same task.

aginzberg:
a) lgtm. b) see comment above - also regarding interface do you think we should allow boxes to be checked in both inprogress and returned simultaneously? i know you handle it correctly so that when you click the icon only the ones in the appropriate section are added to the queue, but maybe if you check some in inprogress and then start checking some more in returned, the ones in inprogress will become unchecked..just a thought...otherwise b lgtm. thanks shirish!

shirishgoyal:
For part b) I will look into the refresh part if it is causing problem
For simultaneous progress, my take is this  - if user is clicking Finish tasks on a specific list, we should show only those tasks to him. So we ignore if he selects tasks in other table, which looks quite natural what user might expect.

shirishgoyal:
Another way to handle it is to maintain one queue and show to user that he is adding tasks to that queue - like 2 tasks is current queue - and then he can work on it together.  We maintain one set of actions for the queue - this will remain only on the frontend side and Queue will be visible only if user adds tasks to it.

aginzberg:
that was kind of how i was originally envisioning the queue, but i dont have a strong preference. @dmorina @nistala @neilthemathguy, thoughts?

shirishgoyal:
In this approach, instead of checkbox, we can have + buttons to add them to queue and user just work on queue always.

static/js/task/controllers/task.controller.js
shirishgoyal: I checked this and it looks fine to me. Will need more elaborate comment on how it is breaking as you might be expecting something else to happen.

My understanding is this - 
a) User can select tasks and click on Finish Tasks, which clears the queue and adds selected tasks to queue
b) User can  work on queue as collection of tasks in order
c) In case user jumps to a different view when queue is not empty (without refresh), queue is still in place but user has no way to go back. So user will have to reselect tasks to continue. (This section can be improved where we allow worker to see the queue and restart from where he left).
d) There is a limit on number of tasks worker can save and keep in progress state.

dmorina:
lgtm as is for now, we have to link this to backend later to maintain state to handle browser refresh but we will do later

static/js/task/controllers/task.controller.js
aginzberg: sorry if this was unclear. to replicate: start the queue with some tasks in it, midway navigate to the task feed, attempt a task. i get this error:
screen shot 2015-09-06 at 3 31 15 pm

message me on slack if this is still unclear.

shirishgoyal: Executed the same steps but no such issue came up. 
The issue you are getting is because of incorrect id being used to fetch task. See new line 38 which corrects this.

aginzberg:
thanks for making that change. lgtm

PR #503, opened 2015-09-09
fix ssl redirect middleware

### LINKED ISSUE #502: SSL Redirect Middleware missing
need to add 'django.middleware.security.SecurityMiddleware', to middleware
otherwise the redirect flags have no effect

dmorina:
resolves #502


nistala:
lgtm

Issue #518, opened 2015-09-12
Ranking Issue


dmorina:
Worker can rank requester multiple times for the same Milestone, if ranking exists we need to update it not create a new one

aginzberg:
#530

PR #530, opened 2015-09-13
Resolves issue #518 creating new rankings rather than updating.

aginzberg:
creates rating_id after submit so that frontend knows to send update when appropriate.


dmorina:
lgtm++

Issue #527, opened 2015-09-13
Task Feed Change for PrototypeTask Evaluation  


neilthemathguy:
We will recruit 20 workers.  
Treatment 1: Revised Task (after prototype task) - we will randomly assign 10 workers.  
Treatment 2: Non-revised (prototype task) - we will randomly assign 10 workers.  

Task Feed changes: Participants in treatment 1 only see the revised tasks
Task Feed changes: Participants in treatment 2 only see the non-revised tasks (original prototype task)

PR #531, opened 2015-09-13
Experiments Infra

dmorina:
adds experimental models

#527 #533 
fixes entry duplicates on task feed
removes cascading at the moment until after CHI


neilthemathguy:
lgtm.

PR #535, opened 2015-09-14
Increase max length, add description to task

shirishgoyal:


crowdsourcing/models.py
```diff
@@ -208,7 +208,7 @@ class Module(models.Model):
             -repetition: number of times a task needs to be performed
     """
     name = models.CharField(max_length=128, error_messages={'required': "Please enter the module name!"})
-    description = models.TextField(error_messages={'required': "Please enter the module description!"})
+    description = models.TextField(error_messages={'required': "Please enter the module description!"}, max_length=2048)
```
dmorina: default VARCHAR is much bigger that this no need to limit it

dmorina:
lgtm

crowdsourcing/models.py
shirishgoyal: It is always better to explicitly mention the length, makes debugging easier.

neilthemathguy:
lgtm

PR #538, opened 2015-09-14
Boomerang requester sampling only

aginzberg:
Starts us off with creating the submodule model and a first pass at the sampling. Now only returns the results if sufficient time has passed since submodule was created.


dmorina:
Store sampling data

dmorina:
lgtm

PR #541, opened 2015-09-14
Boomerang requester frontend

aginzberg:
shows sampled taskworkers and downloads sampled taskworkers when appropriate.


dmorina:
lgtm

PR #544, opened 2015-09-15
send them to task feed upon login

aginzberg:


dmorina:
deployment pending until workers leave system

PR #572, opened 2015-10-01
CHI cleanup

dmorina:
removing all experimental code


aginzberg:
were you intending to leave experimental.py and its various references in?
also we should ditch get_task_workers_sampled and get_task_workers_for_download from serializers/task.py

dmorina:
Not really, can you point where those are? I thought I removed all.

Will remove sampled methods

dmorina:
Oh you mean serializer and viewset, gotcha thanks

PR #574, opened 2015-10-17
Task Authoring and Monitoring Improvements

shirishgoyal:
**CSV**
a) CSV data sources can be added anywhere inside content of label or value of a component
image

b) CSV items can be copied to clipboard on clicking and can be pasted to avoid typing

**Design Canvas**
a) Components can be added without drag/drop
image

b) Same component can be used for display or editing inspired from Google Forms
image

c) Each component (except Instructions) has a label and options/placeholder. Label can be directly associated to Answer provided by worker

d) Instructions allow rich text/html content for better formatting
image

e) Updated Preview for above changes to template compiler and CSV

f) Consolidated Labeled Input component and Text Input into single component
image

**Task**
a) Updated task view for above changes to template compiler and CSV

**Task Monitoring**
a) Replaced display items with Label values from Components in Task


shirishgoyal:
WIP: 
a) Checkbox - handle worker response
b) Handle html entities in CSV headers

bower.json
```diff
@@ -6,20 +6,23 @@
...
-    "angular-ui-sortable": "~0.13.4",
     "jquery": "2.1.4",
     "json3": "~3.2.4",
     "ng-file-upload": "~5.0.9",
     "ng-grid": "2.0.14",
     "ngFx": "~1.1.0",
-    "underscore": "1.8.3"
+    "underscore": "1.8.3",
+    "ckeditor": "~4.5.4",
+    "angular-ckeditor": "~1.0.0"
```
dmorina: ckeditor breaks design, if we are gonna have a rich text editor (which we dont need atm) we are gonna implement it using MD

dmorina:
drop a) and b)
-bring back autocomplete for data source
-drop d) rich text
-add item should stay at the bottom not covering the existing items
-on hover on an item it should show at least Edit, Duplicate, Delete quick actions
-on edit mode use all space for properties

others might follow after iteration

aginzberg:
appearing a bit funky on my computer. I did a bower install so im not sure what is up. is it not supported on chrome? and it doesnt quite look right on firefox. do i need to do something else? see the screenshots below.

Chrome:
screen shot 2015-11-04 at 11 04 08 am

Firefox:
screen shot 2015-11-04 at 11 04 36 am

aginzberg:
More suggested changes
1. Be able to navigate by clicking inside of a template item.
2. Text like "Add question here" should be a placeholder rather than actual text in the label.
3. Click or double-click to edit template item.
4. Unable to delete first element.
5. Do not allow Text fields, radio buttons, etc be editable when not in edit mode.

shirishgoyal:
Task Monitoring Interface
1. Merged Input with Answer
2. Task row can expand to show task detail
3. Rating merged with task row
4. Feedback merged into task detail
5. Bookmark improved
6. Rating Instructions merged and moved to top

image

shirishgoyal:
Updated Task Design Interface

image

crowdsourcing/models.py
```diff
@@ -274,7 +274,7 @@ class TemplateItem(models.Model):
     data_source = models.CharField(max_length=256, null=True)
     layout = models.CharField(max_length=16, default='column')
     type = models.CharField(max_length=16)
-    sub_type = models.CharField(max_length=16)
```
dmorina: we want to keep subtype, will be useful for example for groups of items e.g type: input sub_type: number, but additional changes need to be made so it's fine for now

dmorina:
tried submitting a project 
I get {"modules":[{"template":[{"template_items":[{"values":["This field may not be blank."]},{"values":["This
 field may not be blank."]},{},{}]}]}]}, template looks as follows

dmorina:
screen shot 2015-11-06 at 11 10 27 am

shirishgoyal:
Modified the instructions component to use label instead of values so that we can use values solely for passing options or input values which need worker's input. Instructions is merely a display component so makes it easy to filter at other places.

Added a default text for values for Instructions.

crowdsourcing/models.py
shirishgoyal: I would lean towards adding new component - number than adding specific details to input component. Number can be integer, decimal, range etc. so better to keep it separate.

dmorina: we map 1:1 to html so when I say I want a number input field I cannot specify the datatype other than number, or email or text otherwise the logic becomes more complicated unnecessarily but this comes later no worries

dmorina:
@shirishgoyal lgtm, could you just merge carefully the tabindex changes and then we are good to go

shirishgoyal:
Updated tab support in task display interface

For radio-buttons in material design, tab support for options added instead of default radio-group. 
Instead of spacebar, arrow keys work for selection of choice. Please see comments here: https://github.com/angular/material/issues/3954

dmorina:
lgtm

aginzberg:
for rating on the task monitoring, I think we want to leave it as it was because we aren't rating each individual piece of work done but the worker holistically. i will raise another pr for refining this to only show a few (chosen intelligently) and support search. also we can move the blue box at the top back over to the side rating thing.

shirishgoyal:
@aginzberg: the rating logic remains the same. It happens at individual level. It is only shown once per task row.

dmorina:
The thing is that you are not rating tasks and we dont wanna show more than 5 workers to rate, that's why it used to be separate, these two things shouldn't be mixed, main purpose of this page is monitoring progress and results, in addition requesters have the option to rate their top 5 workers or others too

static/js/task/controllers/task.controller.js
```diff
@@ -58,16 +58,7 @@
...
-                        var tabIndex;
-                        if(obj.role === "display"){
-                            tabIndex = -1;
-                        }
-                        else {
-                            tabIndex = $filter('filter')(self.taskData.task_template.template_items.slice(0, index), {'role': 'input'}).length + 1;
-                        }
-                        angular.extend(obj, {tabIndex: tabIndex});
-                    });
+
```
dmorina: reason for reverting this one?

dmorina:
@shirishgoyal tabindex changes have been lost somewhere?! was that intentional? if not can you bring them back

static/js/task/controllers/task.controller.js
shirishgoyal: We only add tabIndex to components which need to be excluded from tabs. The rest of the components follow the default indexing which works as intended with above code.

PR #581, opened 2015-11-04
added middleware to record login

aginzberg:


dmorina:
add column to User model

dmorina:
lgtm

dmorina:
@aginzberg are we missing a migration here?

aginzberg:
yes. leaving it out to avoid merge issues with the query migrations

dmorina:
ok then

crowdsourcing/middleware/active.py
```diff
@@ -0,0 +1,10 @@
+import django.contrib.auth
+from crowdsourcing.models import UserProfile
+from datetime import datetime
+
+class CustomActiveViewMiddleware():
+	def process_view(self, request, callback, callback_args, callback_kwargs):
+		if not request.user.is_anonymous() and request.path == '/':
```
dmorina: we have to extend this in the future to capture also other paths in case they navigate directly to those paths, but for now good to go

aginzberg: fair point.

PR #582, opened 2015-11-05
Color palette

aginzberg:
Slight change of wording to the feedback section of the task.

Proposed custom gray palette based on the button shadow color discussed today. Other hues derived from some hex arithmetic based on Google MD's specs. This palette allows focus on buttons so tabbing through daemo now has a reasonable user interface.

### DIFF
--- a/static/css/_base.scss
+++ b/static/css/_base.scss
@@ -161,7 +161,3 @@ md-dialog.centered-dialog{
     background-color: #E3F2FD;
   }
 }
-
-.special:not([disabled]):focus {
-  background-color: rgba(158,158,158,0.2);
-}
--- a/static/css/_base.scss
+++ b/static/css/_base.scss
@@ -161,3 +161,7 @@ md-dialog.centered-dialog{
     background-color: #E3F2FD;
   }
 }
+
+.special:focus {
+  font-weight:bold;
+}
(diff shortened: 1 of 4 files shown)


mbernst:
n00b question: what is MD's recommended tab focus interaction? Seems weird
that we'd need to invent our own solution.

aginzberg:
Sorry for the confusion. Previously we had overridden the gray palette to make our own custom gray. But in doing so we had set a few of the hues to white which shouldn't have been white but instead should be some shade of gray. MD by default then uses a specific hue from this custom palette as the button focus. This hue we had erroneously set to white, so there appeared to be no focus at all.

Here, I just flush out all of the gray hues in line with https://www.google.com/design/spec/style/color.html so that when MD tries to grab the correct hue on focus, it actually gets one that's not white.

mbernst:
Ah! OK.

dmorina:
@mbernst there is no spec about the actual keyboard tabbing but it should be the same as normal Tabs or buttons (on hover) that's what we are doing now

dmorina:
lgtm

dmorina:
seems to be a 14% background-color property

aginzberg:
returning to default gray. not sure why we needed a custom one to begin with.

dmorina:
looking good

PR #583, opened 2015-11-05
Card redesign

aginzberg:


static/css/_base.scss
```diff
@@ -160,4 +160,12 @@ md-dialog.centered-dialog{
   .active {
     background-color: #E3F2FD;
   }
+}
+span.inset {
+  padding:5px;
+}
+
+.no-surround {
+  margin: 0px !important;
```
dmorina: it's highly recommended not to use !important unless you really have to

mbernst: An alternative to !important is to pay attention to the CSS import order,
or use selectors that are higher precedence (e.g., more selective).

dmorina: yep +1

aginzberg: sure thing. removed important just now.

dmorina:
lgtm

PR #585, opened 2015-11-12
feature: paypal and payments backend 

### LINKED ISSUE #586: add PayPal API

dmorina:
-adds backend support for paypal and credit card payments
-squashed and optimized migrations
-hashids requirement added for future use

ps. load fixture dmorinaSystemFinancial.json
to test create a new requester account -> that will create the financial accounts, and create a sandbox account with PayPal

closes #586


mbernst:
Wow!

crowdsourcing/models.py
```diff
@@ -559,3 +559,36 @@ class TaskComment(models.Model):
     task = models.ForeignKey(Task, related_name='taskcomment_task')
     comment = models.ForeignKey(Comment, related_name='taskcomment_comment')
     deleted = models.BooleanField(default=False)
+
+
+class FinancialAccount(models.Model):
+    owner = models.ForeignKey(UserProfile, related_name='account_owner', null=True)
```
shirishgoyal: related_name should be named as plural of Class for ForeignKey - In this case, it should be financial_accounts. So it will eventually be used as owner.financial_accounts. It also gives indication that it is 1 to many connection while coding.

crowdsourcing/models.py
```diff
@@ -559,3 +559,36 @@ class TaskComment(models.Model):
...
+    type = models.CharField(max_length=16, default='general')
```
shirishgoyal: Should we stick to choices rather than saving string values ?

crowdsourcing/models.py
```diff
@@ -559,3 +559,36 @@ class TaskComment(models.Model):
...
+    is_active = models.BooleanField(default=True)
+    balance = models.FloatField(default=0)
```
shirishgoyal: Currency value should be DecimalField rather than FloatField with clear distinction for decimal places.

crowdsourcing/models.py
```diff
@@ -559,3 +559,36 @@ class TaskComment(models.Model):
...
+class Transaction(models.Model):
+    amount = models.FloatField()
+    state = models.CharField(max_length=16, default='created')
+    method = models.CharField(max_length=16, default='paypal')
+    sender_type = models.CharField(max_length=8, default='self')
+    sender = models.ForeignKey(FinancialAccount, related_name='transaction_sender')
+    recipient = models.ForeignKey(FinancialAccount, related_name='transaction_recipient')
+    reference = models.CharField(max_length=256, null=True)
+    currency = models.CharField(max_length=4, default='USD')
+    is_deleted = models.BooleanField(default=False)
```
shirishgoyal: Transaction should never be allowed to delete (even archive). We reverse by executing a inverse debit/credit transaction.

crowdsourcing/serializers/payment.py
```diff
@@ -0,0 +1,162 @@
...
+
+class TransactionSerializer(DynamicFieldsModelSerializer):
+    class Meta:
+        model = Transaction
+        fields = ('id', 'sender_type', 'amount', 'currency', 'state', 'method',
+                  'sender', 'recipient', 'reference', 'created_timestamp', 'last_updated')
+        read_only_fields = ('created_timestamp', 'last_updated')
+
+    def create(self, *args, **kwargs):
+        transaction = Transaction.objects.create(**self.validated_data)
```
shirishgoyal: Avoid passing values directly from API. This will otherwise allow making transaction between any two users. We should be restricting between logged in user and worker only (avoid passing worker id from API too as it should be available from task itself).

crowdsourcing/serializers/payment.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+class TransactionSerializer(DynamicFieldsModelSerializer):
+    class Meta:
+        model = Transaction
+        fields = ('id', 'sender_type', 'amount', 'currency', 'state', 'method',
+                  'sender', 'recipient', 'reference', 'created_timestamp', 'last_updated')
+        read_only_fields = ('created_timestamp', 'last_updated')
+
+    def create(self, *args, **kwargs):
```
shirishgoyal: Use atomic transactions for exposed API for payments. 
https://docs.djangoproject.com/en/1.8/topics/db/transactions/

crowdsourcing/viewsets/payment.py
```diff
@@ -0,0 +1,46 @@
...
+        serializer = PayPalPaymentSerializer(data=request.data, context={"request": request})
+
+        if serializer.is_valid():
+            flow_data, http_status = serializer.create()
+            return Response(data=flow_data, status=http_status)
+        else:
+            return Response(data=serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+    @list_route(methods=['post'])
+    def execute(self, request, *args, **kwargs):
```
shirishgoyal: Use atomic transactions for exposed API for payments. 
https://docs.djangoproject.com/en/1.8/topics/db/transactions/

crowdsourcing/migrations/0001_initial.py
```diff
@@ -48,6 +48,12 @@ class Migration(migrations.Migration):
             ],
```
shirishgoyal: Only migrations which have gone into production should have been merged.

crowdsourcing/models.py
dmorina: it's suggested for many to many relationships else it's main purpose is to have unique constraint names, but surely this related_name was not chosen wisely :D

crowdsourcing/models.py
dmorina: whichever is fine

crowdsourcing/viewsets/payment.py
dmorina: how does that help here? actually in this case it's the opposite you dont want to tie the transaction commit/rollback to the result of the whole view code block, if user hits this endpoint we want to make sure that we capture this data and commit to DB 

maybe you can elaborate what you had in mind here

shirishgoyal: Since API is interacting with 3rd party API, request timeout can happen anytime and it is much easier to just revert the whole transaction on 3rd party system than investigating the side effects and executing steps manually.

dmorina: if you have a look at  crowdsourcing/serializers/payment.py line 52 you see that there is no new transaction unless the call returns successfully :) so if that fails there is nothing to revert

crowdsourcing/serializers/payment.py
```diff
@@ -0,0 +1,162 @@
...
+        transaction = Transaction.objects.create(**self.validated_data)
+        transaction.recipient.balance += transaction.amount
+        transaction.recipient.save()
+        if not transaction.sender.is_system:
```
aginzberg: This will lead to massive inflation if we never reduce the system account's balance.

crowdsourcing/serializers/payment.py
```diff
@@ -0,0 +1,162 @@
...
+                "state": "approved",
+                "method": payment["payer"]["payment_method"],
+                "sender": FinancialAccount.objects.get(is_system=True, type="paypal_deposit").id
+            }
+            if not self.context['request'].user.is_anonymous():
+                transaction["sender_type"] = "self"
+            else:
+                transaction["sender_type"] = "other"
+            serializer = TransactionSerializer(data=transaction)
+            if serializer.is_valid():
```
aginzberg: If this fails for some reason, we need to make sure we end up crediting the requester's account later.

crowdsourcing/serializers/payment.py
```diff
@@ -0,0 +1,162 @@
...
+        else:
+            return payment.error['message'], status.HTTP_400_BAD_REQUEST
+
+
+class CreditCardSerializer(serializers.Serializer):
+    type = serializers.ChoiceField(choices=['visa', 'mastercard', 'discover', 'american_express'])
+    number = serializers.CharField(min_length=13, max_length=19)
+    expire_month = serializers.IntegerField(min_value=1, max_value=12)
+    expire_year = serializers.IntegerField()
+    cvv2 = serializers.IntegerField(min_value=111, max_value=9999)
```
aginzberg: I believe this can start with a 0

crowdsourcing/serializers/payment.py
```diff
@@ -0,0 +1,162 @@
...
+            ConditionallyRequiredValidator(field='method', field2='credit_card', value='credit_card')
+        ]
+
+    def build_payment(self, *args, **kwargs):
+        payment = {"intent": "sale",
+                   "payer": {
+                       "payment_method": self.validated_data['method']
+                   },
+                   "redirect_urls": {
+                       "return_url": "http://localhost:8000/paypal-success",
```
aginzberg: These should be relative

crowdsourcing/serializers/payment.py
```diff
@@ -0,0 +1,162 @@
...
+    class Meta:
+        validators = [
+            InequalityValidator(
+                field='amount', operator='gt', value=1
+            ),
+            ConditionallyRequiredValidator(field='method', field2='credit_card', value='credit_card')
+        ]
+
+    def build_payment(self, *args, **kwargs):
+        payment = {"intent": "sale",
```
aginzberg: At some point, we should also let people cash out.

crowdsourcing/serializers/payment.py
```diff
@@ -0,0 +1,170 @@
...
+            return 'Payment executed successfully', status.HTTP_201_CREATED
+        return serializer.errors, status.HTTP_400_BAD_REQUEST
+
+
+class CreditCardSerializer(serializers.Serializer):
+    type = serializers.ChoiceField(choices=['visa', 'mastercard', 'discover', 'american_express'])
+    number = serializers.CharField(min_length=13, max_length=19)
+    expire_month = serializers.IntegerField(min_value=1, max_value=12)
+    expire_year = serializers.IntegerField()
+    cvv2 = serializers.RegexField(regex='^[0-9]{3,4}$', max_length=3, min_length=3, required=True)
```
aginzberg: max_length = 4?

crowdsourcing/serializers/payment.py
```diff
@@ -0,0 +1,170 @@
...
+        model = Transaction
+        fields = ('id', 'sender_type', 'amount', 'currency', 'state', 'method',
+                  'sender', 'recipient', 'reference', 'created_timestamp', 'last_updated')
+        read_only_fields = ('created_timestamp', 'last_updated')
+
+    def create(self, *args, **kwargs):
+        transaction = Transaction.objects.create(**self.validated_data)
+        transaction.recipient.balance += transaction.amount
+        transaction.recipient.save()
+        if not transaction.sender.is_system and transaction.sender.type != 'paypal_external':
```
aginzberg: or instead of and?

crowdsourcing/serializers/payment.py
```diff
@@ -0,0 +1,177 @@
...
+            recipient_profile = get_model_or_none(UserProfile, user__username=self.validated_data['username'])
+        recipient = get_model_or_none(FinancialAccount, owner=recipient_profile, type='requester')
+        paypalbackend = PayPalBackend()
+        payment = paypalbackend.paypalrestsdk.Payment(payment_data)
+        if payment.create():
+            redirect_url = next((link for link in payment['links'] if link['method'] == 'REDIRECT'), '#')
+            flow_data = {
+                "redirect_url": redirect_url['href'],
+                "paypal_id": payment.id
+            }
```
aginzberg: I think you're crashing here when the method is a credit card because I believe you aren't getting back a redirect_url and so redirect_url['href'] breaks.

crowdsourcing/serializers/payment.py
```diff
@@ -0,0 +1,177 @@
...
+    cvv2 = serializers.RegexField(regex='^[0-9]{3,4}$', required=True)
+    first_name = serializers.CharField()
+    last_name = serializers.CharField()
+
+    class Meta:
+        fields = ('type', 'number', 'expire_month', 'expire_year', 'cvv2', 'first_name', 'last_name')
+
+
+class PayPalPaymentSerializer(serializers.Serializer):
+    amount = serializers.IntegerField()
```
aginzberg: This shouldn't be an integer I don't think

dmorina: doesnt it accept only integer values? :D

aginzberg: yeah, but don't you want people to be able to send non-integer values?

dmorina: so we round it for them?

aginzberg: no i would imagine paypal accepts non-integer values lol

dmorina: lol, misunderstanding, all clear :D

PR #592, opened 2015-11-17
task feed sorting

dmorina:
-adds user preferences including task feed options


dmorina:
will include in task feed redesign

PR #599, opened 2015-11-24
Milestone and Project field changes

dmorina:
-allows nulls and blanks for few fields, defaults are in place
-module validation occurs if status=3 (published) note that this wont be called on queryset.update


dmorina:
@aginzberg this one just allows nulls and adds defaults on few fields, merging to continue with authoring

PR #600, opened 2015-11-30
moving to new design

dmorina:
-initial changes towards Chris' designs  {Font: Fira Sans, Option 1 from proposal see #design channel}
-default themes switched for the moment to purple and light-blue until we have our own palettes


aginzberg:
looks good for now. lgtm

PR #601, opened 2015-12-10
new authoring + upgrades

dmorina:
- implements new authoring design
- upgrade to django 1.9
- migration fixes
- requester files extended and renamed to BatchFiles (upload and remove supported)


crowdsourcing/migrations/0000_get_requester_ratings_fn.py
```diff
@@ -0,0 +1,67 @@
...
+                                             wrr.target_id,
+                                             wrr.weight
+                                           FROM crowdsourcing_workerrequesterrating wrr
+                                             INNER JOIN (
+                                                          SELECT
+                                                            origin_id,
+                                                            target_id,
+                                                            MAX(last_updated) AS max_date
+                                                          FROM crowdsourcing_workerrequesterrating
+                                                          GROUP BY origin_id, target_id
```
aginzberg: Filter before the join? Same in the other function

dmorina: it doesnt matter in this case

aginzberg:
lgtm. really nice work. i think we can merge here, and refine design and other issues / missing features as we go.

PR #603, opened 2015-12-16
Project actions

aginzberg:
Removes project level from my projects. Allows pausing/resuming, editing/deleting, and forking of projects. These actions are only allowable depending on the status of the project.

Note: when md-secondary comes out we can make the icons overlay the highlighted list item better and avoid all of the stopPropagation junk


dmorina:
grrrr ... need to update sql migrations to status=2

aginzberg:
i dont think so. status 2 is used only temporarily when the status is first being published, when the tasks need to be created for the first time. status 3 is when the project is in progress and should be displayable on the task feed.

dmorina:
clicking on the buttons (pause/fork) opens the module on Firefox

static/js/project/controllers/my-project.controller.js
```diff
@@ -71,8 +77,77 @@
...
+            Project.update(item.id, {status: 3}, 'module').then(
+                function success(response) {
+                    $mdToast.showSimple('Resumed ' + item.name + '!');
+                    item.status = 3;
+                },
+                function error(response) {
+                    $mdToast.showSimple('Could not resume module.');
+                }
+            ).finally(function () {});
+        }
```
dmorina: please merge resume and pause into one function

dmorina:
lgtm

PR #604, opened 2015-12-17
Remote Content - iframe

shirishgoyal:
image

a) Loads content from remote URL in a sandbox
b) In task mode, appends unique identifier to remote URL for consumption by the remote content  to pass back the data to the system.


shirishgoyal:
Backend
- Each iframe item now has a unique identifier - taskId, taskworkerId and templateItemId
- API endpoint : /api/done/?daemo_id=<identifier>
- Must be a POST request and have some data 
- Will only accept when task is in state of in progress, submission, returned.

mbernst:
Sometimes the external sites will want to record onto all the data
themselves, and so will want to POST back so they worker gets paid but not
add any additional data. Is the extra data necessary?

shirishgoyal:
In current implementation, we have made it mandatory, but we can bypass, if that is the common case.

mbernst:
For iframes, let's not require it. Thanks!

On Saturday, December 19, 2015, Shirish Goyal <notifications@github.com>
wrote:

> In current implementation, we have made it mandatory, but we can bypass,
> if that is the common case.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/crowdresearch/crowdsource-platform/pull/604#issuecomment-165987436>
> .
>


-- 
sent from my NLS

crowdsourcing/viewsets/task.py
```diff
@@ -261,3 +284,37 @@ class CurrencyViewSet(viewsets.ModelViewSet):
 
     queryset = Currency.objects.all()
     serializer_class = CurrencySerializer
+
+
+class ExternalSubmit(APIView):
+    def post(self, request, *args, **kwargs):
+        identifier = request.GET.get('daemo_id', False)
```
dmorina: use query_params
and let's allow ajax calls as well request.POST wont be filled in that case use request.data

csp/settings.py
```diff
@@ -150,8 +148,8 @@
     'compressor.finders.CompressorFinder',
 )
 
-COMPRESS_ENABLED = True
-COMPRESS_OFFLINE = True
+COMPRESS_ENABLED = False
+COMPRESS_OFFLINE = False
```
dmorina: this is overridden in local settings

static/django_templates/base/index.html
```diff
@@ -1,4 +1,4 @@
-{% load staticfiles %}
+{% load staticfiles compress %}
```
dmorina: reason for disabling compressing is because django-compressor/master does not support django 1.9 yet, dev branch is fixed now

crowdsourcing/viewsets/task.py
mbernst: Can you explain the user workflow if they're using AJAX? eg making the
submission in the background?

On Saturday, December 19, 2015, Durim Morina <notifications@github.com>
wrote:

> In crowdsourcing/viewsets/task.py
> <https://github.com/crowdresearch/crowdsource-platform/pull/604#discussion_r48095890>
> :
>
> > @@ -261,3 +284,37 @@ class CurrencyViewSet(viewsets.ModelViewSet):
> >
> >      queryset = Currency.objects.all()
> >      serializer_class = CurrencySerializer
> > +
> > +
> > +class ExternalSubmit(APIView):
> > +    def post(self, request, *args, **kwargs):
> > +        identifier = request.GET.get('daemo_id', False)
>
> use query_params
> and let's allow ajax calls as well request.POST wont be filled in that
> case use request.data
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/crowdresearch/crowdsource-platform/pull/604/files#r48095890>
> .
>


-- 
sent from my NLS

dmorina: if we listen on $location change of the iFrame as we discussed and we receive a submission from their page then we can mark it as submitted, workflow doesn't change from the regular POST
workflow is: 1. they submit the form data whichever way they prefer (ajax or regular POST) and 2. they either let us know the user is `done` by location change to something like `/task-done` and we capture that info and continue from there

static/js/template/services/template.service.js
```diff
@@ -354,88 +355,91 @@
...
+                 data_source: null,
+                 role: 'display',
+                 label: '',
+                 name: "iFrame",
+                 icon: 'web',
+                 type: 'video'
+                 },*/
                 {
-                    tooltip: "External Content (iFrame)",
+                    name: "Remote Content",
```
dmorina: use the new way of writing the templates, I moved them to files because it's getting harder to maintain as these are growing, see media-edit and media templates it's very easy to plug iFrame in all the rest will work out of the box

csp/settings.py
```diff
@@ -104,8 +102,8 @@
                 'django.contrib.messages.context_processors.messages',
             ],
             # 'loaders': [
-            # 'django.template.loaders.filesystem.Loader',
-            # 'django.template.loaders.app_directories.Loader',
+            #     'django.template.loaders.filesystem.Loader',
```
dmorina: irrelevant but if you follow PEP8 there must be only 1 space between # and the comment body

static/django_templates/base/index.html
dmorina: or stable version 1.6 with django 1.9 support can be pip installed from my fork

crowdsourcing/viewsets/task.py
mbernst: What if they send multiple AJAX requests? Do we append to the data we
record? Or something else?

On Saturday, December 19, 2015, Durim Morina <notifications@github.com>
wrote:

> In crowdsourcing/viewsets/task.py
> <https://github.com/crowdresearch/crowdsource-platform/pull/604#discussion_r48096791>
> :
>
> > @@ -261,3 +284,37 @@ class CurrencyViewSet(viewsets.ModelViewSet):
> >
> >      queryset = Currency.objects.all()
> >      serializer_class = CurrencySerializer
> > +
> > +
> > +class ExternalSubmit(APIView):
> > +    def post(self, request, *args, **kwargs):
> > +        identifier = request.GET.get('daemo_id', False)
>
> if we listen on $location change of the iFrame as we discussed and we
> receive a submission from their page then we can mark it as submitted,
> workflow doesn't change from the regular POST
> workflow is: 1. they submit the form data whichever way they prefer (ajax
> or regular POST) and 2. they either let us know the user is done by
> location change to something like /task-done and we capture that info and
> continue from there
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/crowdresearch/crowdsource-platform/pull/604/files#r48096791>
> .
>


-- 
sent from my NLS

dmorina: is that a case when it's a multi step task or something? the results field can be anything list/object so Yes in multiple requests we store them as a list, if we know it's a re-submit then we update the existing one, this is independent of whether it's an Ajax call or regular form POST (there can be multiple forms)

static/django_templates/base/index.html
shirishgoyal: I intentionally wanted to disable compressor from settings for project if we are not using it rather than removing compress everywhere. 

If you have a fork, we can enable it again.

static/templates/template/components/iframe-edit.html
```diff
@@ -0,0 +1,30 @@
+<ng-include src="'/static/templates/template/components/question-edit.html'"></ng-include>
+<div layout="row" layout-align="start center">
+    <iframe class="remote-content" ng-src="{{instance.getTrustedUrl(item.aux_attributes.src)}}" frameborder="0" width="100%"
```
dmorina: if this is the only unique snippet you could merge with media-edit (maybe rename) and use ng-if there

static/django_templates/base/index.html
dmorina: For it to work you either have to set `APP_DIRS: True` or specify the `loaders` app_directories loader to be exact

shirishgoyal: Fixed compressor and tested on heroku

crowdsourcing/viewsets/task.py
```diff
@@ -259,3 +282,34 @@ class CurrencyViewSet(viewsets.ModelViewSet):
...
+            task_worker_id, task_id, template_item_id = hash.decode(identifier)
+
+            with transaction.atomic():
+                task_worker = TaskWorker.objects.get(id=task_worker_id, task_id=task_id)
+                task_worker_result, created = TaskWorkerResult.objects.get_or_create(task_worker_id=task_worker.id,
+                                                                                     template_item_id=template_item_id)
+                # only accept in progress, submitted, or returned tasks
+                if task_worker.task_status in [1, 2, 5]:
+                    task_worker_result.status = 1
+                    task_worker_result.result = request.data
```
dmorina: how do you want to handle the point made by @mbernst on multiple submits  (i.e multiple forms or multiple ajax calls)?

shirishgoyal: I thought you already replied to that :)

Currently, multiple submits will overwrite the data if task is still in progress.

We expect maximum one form in an iframe but there is no technical limitation as such and we leave it to the author on how he wants to use it.

For a multi-step form, which can push data in partial mode, we can probably allow author to add another url parameter like - append - to signify data has to be appended to existing data. But we still expect component key names in form to be different in such case.

For all such extra options, we will need to instruct author somewhere on how these parameters can be used.

dmorina:
lgtm

aginzberg:
lgtm

PR #613, opened 2015-12-27
Refactoring + Payment FE + Contributors FE

shirishgoyal:
### Refactoring
- Model choices used in pythonic way
- Replace all direct usages of choices based options with canonical style
- Merge FE templates to their respective modules

### Bugs
- Remove flicker in task feed (available tasks)
- Enable Drag Drop for Template Design

### Payment Frontend
- Allow user to deposit from profile page via Paypal/Credit Card
- Interact with Paypal services - Success/Cancel routes

image

### Contributors Frontend
- Authors List 
- Active Contributors List - Alphabetical
- Data (to be updated)

image


crowdsourcing/models.py
```diff
@@ -211,18 +217,20 @@ class Project(models.Model):
...
-                (4, 'Completed'),
-                (5, 'Paused')
-                )
-    permission_types = ((1, "Others:Read+Write::Workers:Read+Write"),
-                        (2, 'Others:Read::Workers:Read+Write'),
-                        (3, 'Others:Read::Workers:Read'),
-                        (4, 'Others:None::Workers:Read')
-                        )
-    status = models.IntegerField(choices=statuses, default=1)
+    STATUS = Choices(
```
dmorina: nice!

crowdsourcing/serializers/payment.py
```diff
@@ -64,6 +66,9 @@ def create_transaction(self, payment, payer_id):
         flow.state = 'approved'
         flow.payer_id = payer_id
         flow.save()
+
+        sender, created = FinancialAccount.objects.get_or_create(is_system=True, type="paypal_deposit")
```
dmorina: this does not belong here, will be moved to a migration

dmorina:
html templates should remain under `templates` folder (nobody expects them to be under `js`);
there are two main approaches to this: either keep them separate with same folder structure OR under `static` you have one folder per component and inside it you have js+html+scss

crowdsourcing/models.py
```diff
@@ -1,13 +1,15 @@
 from django.contrib.auth.models import User
 from django.db import models
+from model_utils import Choices
```
dmorina: it is the right way to handle Choices although Im not convinced that it's worth adding a dependency
same thing can be achieved without it 
```
class ModelEg(Model):
   PROJECT_CREATED = 1 
   PROJECT_PUBLISHED = 2
   statuses = ((PROJECT_CREATED, 'Created'),
                (PROJECT_PUBLISHED, 'Published'), )

```
or is there something more to it?

shirishgoyal: We can always achieve everything without dependencies. We decide whether we want to use a simpler structure without reinventing it.

shirishgoyal:
I am open to renaming "js" folder as "components". We will anyways be moving to this structure when upgrading to Angular 2.0 in near future. Also it is much easier to find templates when we are working on that module.

dmorina:
We need to discuss that, for now let's keep it AS-IS

dmorina:
Alright, let's keep the templates as it is now, there is no strong argument to change it

crowdsourcing/models.py
dmorina: true; we always have to check what makes sense to write by ourselves and what not, we have to see what functionality they bring to us, for this particular case since these statuses are being used across the app we create a python module `status.py` and import it as `from crowdsourcing import status` and use it in Models, Serializers, Viewsets wherever; the constants can follow a certain pattern  MODELNAME_STATUS -> e.g PROJECT_PUBLISHED

crowdsourcing/serializers/payment.py
dmorina: @shirish.goyal could you check this, everything else is well

crowdsourcing/migrations/0012_auto_20160107_0521.py
```diff
@@ -0,0 +1,22 @@
...
+    account.objects.get_or_create(is_system=True, type="paypal_deposit")
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ('crowdsourcing', '0011_auto_20151221_1618'),
+    ]
+
+    operations = [
+        migrations.RunPython(create_system_financial_account),
```
dmorina: TypeError: create_system_financial_account() takes exactly 1 argument (2 given)

dmorina:
@shirishgoyal when depositing money from the dialog (i select Credit Card and then change it to PayPal) could you delete credit_card information on Request (this is a bug on the backend too, apologies for that)

dmorina:
screen shot 2016-01-08 at 4 33 57 pm

Strange bug (im on FF): after I change the position of an item I cannot edit the Question text anymore, if I just keep typing it keeps prepending it

shirishgoyal:
- Fixed migration
- Remove credit_card when paypal is chosen
- Fix handle and style changes

image

dmorina:
All is well

PR #620, opened 2016-01-04
Environ definition

shirishgoyal:
- Add local requirements with additional pip dependencies for developer machines
- Add environ variables for Registration
- Update heroku documentation

We will need to transfer all API Keys and Secret values to environ to avoid sharing them on git.


dmorina:
lgtm

Issue #617, opened 2015-12-28
task feed redesign


dmorina:
-redesign according to Michael's and Christopher's final design

PR #624, opened 2016-01-10
Pretty taskfeed

aginzberg:
addresses #617 
screen shot 2016-01-10 at 2 23 19 pm
screen shot 2016-01-10 at 2 23 36 pm


static/templates/task-feed/main.html
```diff
@@ -1,106 +1,71 @@
...
+    <div ng-hide="!project.available_tasks" ng-repeat="project in taskfeed.projects" class="md-whiteframe-z2">
+        <div class="_task-top" ng-click="taskfeed.showPreview(project)">
+          <div layout="row">
+            <div flex="10" class="_price">
+              {{ project.price | currency:"$":2}}
+            </div>
+            <div flex="70" class="_title">
+              {{ project.name }}
+            </div>
+            <div flex="20" class="_author _right-offset" ng-click="taskfeed.openRequesterProfile(project.owner)">
```
dmorina: probably we remove ng-click for the moment, just makes the thing reload

dmorina:
i think we didnt want to show Minutes for prototype tasks at least, and do a text-transform: uppercase on that field

dmorina:
and when there is no rating do we show it as a 2 or nothing? might be confusing @mbernst 

i think later we want to differentiate between avg rating and my rating for this requester probably 2 colors

static/templates/task-feed/main.html
```diff
@@ -1,106 +1,71 @@
...
-            <span>
-                <span class="inset">
-                    <md-icon md-font-set="material-icons">attach_money</md-icon>
-                    USD {{ project.price }}
-                </span>
-                <span class="inset">
-                    <md-icon md-font-set="material-icons">list</md-icon>
-                    Tasks (available/total): {{ project.available_tasks + '/'+project.total_tasks}}
+  <div class="_feed-list">
+    <div ng-hide="!project.available_tasks" ng-repeat="project in taskfeed.projects" class="md-whiteframe-z2">
```
dmorina: z2 seems a bit too much, we probably have to go for a custom one that matches Chris' design

mbernst:
The norm (e.g., on Netflix) is that the average rating is a lighter color,
and the ones you've explicitly filled in are darker. So you can stick with
one color and do a lighter/darker variant. If there's no rating, go with
the average rating, yeah?

dmorina:
agreed, if average does not exist that means unrated by anyone, hence empty circles?

mbernst:
I thought everyone has a 2.0 average by default, with an exponentially-weighted moving average. So all new requesters should be 2.0.

aginzberg:
Technically 1.99, but it appears as 2.0. We can make new and requesters unrated by this particular worker a lighter color and ones theyve rated darker.

dmorina:
they do get a 1.99 in feed sorting, but if we show it as 1.99 or 2.0 then you will have 3 types of ratings:
1. mine 2. average if mine doesnt exist and 3. 1.99 if neither exists

mbernst:
We'll presumably need to round to the nearest half-point for the interface
anyway. So 1.99 will appear below 2.0 in the feed, but will still appear as
two circles. Does that work?

dmorina:
Yeah i think we are doing sth like that with % though, my point is just that we need let the user know which requester is new and which rated by other workers only

dmorina:
Anyways we can start with two states and we see in the future

mbernst:
Got it. I'm not immediately convinced that the difference between "new" and
"others have rated" is super important. But we could replace the rating
circles with a label that says "new" if we want. (e.g., the angular
equivalent of http://getbootstrap.com/components/#labels)

aginzberg:
changes made to resolve comments

static/js/task-feed/controllers/task-feed.controller.js
```diff
@@ -33,6 +33,8 @@
         self.saveComment = saveComment;
         self.loading = true;
         self.getStatusName = getStatusName;
+        self.openRequesterProfile = openRequesterProfile;
```
dmorina: comment out or delete, otherwise page wont load

dmorina:
and resolve conflicts

dmorina:
looking good

mbernst:
@shirishgoyal we need one more lgtm

shirishgoyal:
Will review it shortly. Thanks for the headsup.

shirishgoyal:
Issues found:
a) Requester can see his/her own task in task-feed and can accept and complete it

b) Worker cannot see a sample of the task. On clicking the task widget / rating  (except Accept button) , keeps on showing error about permission
image

c) Error shows JSON as text

dmorina:
a) is desirable 
b) can you post the error

shirishgoyal:
If a) is needed, please ignore it as issue then.
b) Already posted. See error in screenshot

aginzberg:
good catch @shirishgoyal. permissions should be fixed.

dmorina:
@shirishgoyal is this good to merge?

shirishgoyal:
LGTM

PR #636, opened 2016-01-18
MTurk integration

dmorina:
-mturk integration via ExternalQuestion 
-celery and redis added, need to install redis-server
i set the task frequency to 60s for testing purposes, will be changed to 30m

deployed to https://daemo-test.herokuapp.com

md-accent to be defined (waiting for the palette from Chris), for the moment the light blue accent is applied


mturk/views.py
```diff
@@ -0,0 +1,11 @@
+from django.shortcuts import render
+
+
+def mturk_index(request, *args, **kwargs):
+    from mturk.interface import MTurkProvider
+    from crowdsourcing.models import Project, Task
+    provider = MTurkProvider('https://localhost:8000')
+    project = Project.objects.get(id=1)
```
shirishgoyal: Should be using SITE_HOST instead of localhost
create_external_question uses taskId which is not used here.

dmorina: Apologies, all that was for testing only return render belongs there will clean up

shirishgoyal:
#### Task created on MTurk from Daemo
image

#### Task in preview mode
image

#### Task being worked upon
image

#### Task submission from MTurk available on Daemo
image

#### Task although searchable on MTurk but not available for same worker anymore
image

### Suggested improvements:
a) Allow task to be paused/restarted on Daemo and thus on MTurk (expire/extend)
b) Allow task to be rejected from Daemo to MTurk (atleast in test mode)
c) Task Submission - Task tab should be select only (currently it is toggle)

aginzberg:
really nice job, lgtm (p.s. remove console.log)

PR #639, opened 2016-01-19
Alert a successful registration and to activate account

kaushik94:
screen shot 2016-01-19 at 9 04 43 pm


kaushik94:
@shirishgoyal done

dmorina:
MD uses `toasts` for notifications rather than dialogs, #642 does that; so we will start with toasts and later if designers decide to embed the success message into the form itself we will change it

kaushik94:
Looks better than mine, +1.

PR #640, opened 2016-01-20
Email backend for SendGrid

shirishgoyal:
a) Add a new backend for SendGrid to remove hardcoded dependencies
b) Allow local settings to override email backend to use django's available backends if needed
c) Add console backend as default for local settings

### DIFF
--- a/crowdsourcing/emails.py
+++ b/crowdsourcing/emails.py
@@ -1,43 +1,23 @@
 from django.conf import settings
+from django.core.mail import EmailMultiAlternatives
 
 
-def send_activation_email_gmail(email, host, activation_key):
-    """
-        This sends the activation link to the user using Gmail, the content will be moved to template files
+def send_mail(email_from, email_to, subject, text_content, html_content):
+    mail = EmailMultiAlternatives(
+        subject=subject,
+        body=text_content,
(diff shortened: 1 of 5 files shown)


dmorina:
should we use `sendmail` for dev systems?

dmorina:
lgtm

shirishgoyal:
I think any email provider like console or file based is enough for dev system. 

Apart from that, backend should not limit developer's choice and we leave it to him/her to decide for local system.

crowdsourcing/emails.py
```diff
@@ -84,4 +51,68 @@ def send_password_reset_email(email, host, reset_key):
                    ""<br><p>If you didn't request a password reset please click the following link: <br>" + \
                    '<a href="' + reset_url + '/0' + '">' + reset_url + '/0' + \
                    '</a><br><br> Greetings,<br> <strong>Daemo Team</strong>'
-    send_grid(to, subject, text_content, html_content)
+
+    send_mail(email_from, email_to, subject, text_content, html_content)
+
+# def send_activation_email_gmail(email, host, activation_key):
```
dmorina: you can drop all this below

Issue #641, opened 2016-01-20
Display a notification alert on successful registration


Nalinc:
In present scenario, whenever a user registers successfully on Daemo, he is redirected to login page right away. At this point if a login attempt is made, an error message is displayed saying "Account is not activated yet". 

For better usability, user should be presented with a notification message mentioning that an email with activation link has been sent after successful registration. 

Example of such notifications can be seen in "Forgot Password" workflow and a couple of other places.

Issue #626, opened 2016-01-12
Sign up form not showing "activation needed" message


misaelupg:
After the user signs up, there's no message to inform it about activation needed with link sent to its email.

PR #642, opened 2016-01-20
#641: Display notification message on successful registration.

Nalinc:
After successful registration, user should be presented with a notification message mentioning that an email with activation link has been sent.

This PR let user see the message "Email with an activation link has been sent", when the user is redirected from registration page to login page.

register-alert


dmorina:
tip: in your comments use text like "closes #641" and after merge the issue will be closed automatically :)

Nalinc:
Closes #626 and #641

Issue #500, opened 2015-09-08
Allow user to update his/her profile


shirishgoyal:
Currently user's profile is read-only.
As a user, I should be able to edit my profile information

PR #644, opened 2016-01-25
Issue #500 Engineering-Profile

thunderboltsid:
Added form for modifying user profile info based on the viewset update_profile api. Haven't written tests yet but it works.

screen shot 2016-01-25 at 6 26 05 pm
screen shot 2016-01-25 at 6 26 19 pm


dmorina:
run `flake8 .` before pushing code :)

thunderboltsid:
So, I presume flake8 is some sort of PEP8 enforcing system, or?

dmorina:
yep

crowdsourcing/viewsets/user.py
```diff
@@ -155,3 +158,33 @@ def retrieve(self, request, *args, **kwargs):
         user = get_object_or_404(self.queryset, user=request.user)
         serializer = UserPreferencesSerializer(instance=user)
         return Response(serializer.data)
+
+
+class CountryViewSet(viewsets.ModelViewSet):
```
dmorina: when using a ModelViewset it will inherit 5 default actions -> endpoints 
.list(), .retrieve(), .create(), .update(), and .destroy() that means that you wont need the `get_country` in this particular case, and since you have a serializer_class you can use `Response(data=..., status=....)`

Best is to use mixins and only inherit the actions you need  see `UserViewSet`

crowdsourcing/viewsets/user.py
```diff
@@ -155,3 +158,33 @@ def retrieve(self, request, *args, **kwargs):
...
+    """
+    JSON response for returning cities
+    """
+    lookup_field = "name"
+    queryset = City.objects.all()
+    serializer_class = CitySerializer
+
+
+    @list_route()
+    def get_country(self, request):
```
dmorina: same as above

static/js/user/controllers/user.controller.js
```diff
@@ -10,16 +10,50 @@
...
+        vm.edit=false;
+        vm.genders=[
+            {"key":"M", "value":"Male"},
+            {"key":"F", "value":"Female"},
+            {"key":"O", "value":"Other"}
+        ];
+        vm.countries = [];
+        vm.cities = [];
+
+        $http({
```
dmorina: we interact with the api using services, that way we can re-use the methods anywhere, would be better to move this call to there and call it accordingly from the controller

dmorina:
address should not show as json object

dmorina:
Fix address front end to show names not ids and json, this breaks registration could you check it please

thunderboltsid:
Will do. Already have a patch. will push it soon.

static/templates/user/profile.html
```diff
@@ -4,23 +4,61 @@
...
                 </div>
-                <div layout="column" layout-align="start" class="md-padding" flex>
+                <div ng-hide="vm.edit" layout="column" layout-align="start" class="md-padding" flex>
                     <div class="inset">
-                        <p><span class="md-subheader">Date of Birth :</span> {{ vm.user.birthday || 'Not Available'}}</p>
+                        <p><span class="md-subheader">Date of Birth :</span> {{ vm.user.birthday || 'Not Available'}}
+                        </p>
 
-                        <p><span class="md-subheader">Address :</span> {{ vm.user.address || 'Not Available'}}</p>
+                        <p><span class="md-subheader">Address :</span> {{ vm.user.address.street + ", " + vm.user.address.city_obj.name + ", " + vm.user.address.country_obj.name || 'Not Available'}}</p>
```
aginzberg: If they haven't put in an address yet, it looks kind of funky with just commas.

aginzberg:
Could not replicate @dmorina's issue. One thing is do you want to maybe include a data migration or fixture with some city and country data? Otherwise looks good.

thunderboltsid:
Upvote for the fixture, perhaps we can package on from the existing database and include it as a part of the developer package.

shirishgoyal:
Bugs:
a) Address shows commas when values are not filled in
b) Date of Birth has no label and hence, not apparent what calendar is for
c) No City or Country data to choose from. Fails to update profile.
d) Edit profile is a toggle button which turns into Cancel to shift to view mode (looks less intuitive to me). It should be right next to Submit.

thunderboltsid:
@mbernst mentioned that we should add more fields to the model (UserProfile) - Namely, tags and ethnicity. So, that's what I'll do in the coming days. Exam tomorrow, so will work after that.

mbernst:
Good luck on the exam! Looking forward to seeing you back.

thunderboltsid:
Ok, no clue why the build is failing. Says: No module named django_orm. That's weird.

qwertyone:
I have a concern. Ethnicity data crosses into a legal grey area to
discrimination. Should it be included?

mbernst:
This is a great question. How do we balance scientists' need to report that
data vs. discrimination? Also, how can we prove there is or isn't
discrimination without the data? What if there's tons now that's implicit
(and there certainly is against Indian turkers, for instance) but we can't
demonstrate it because we don't have the information?

Maybe one easy option would be to require requesters to explicitly request
certain demographic variables be shared, and Daemo to alert the worker that
this is happening when they decide to accept a task.

thunderboltsid:
@mbernst I believe that the demographic information is sensitive to the user and should not be shared. Also, allowing the requester to request demographic variables such as ethnicity of the worker is ethically questionable for most set of tasks.

mbernst:
There's a difference between information the *platform* has — e.g., so we can report general demographics, and what each *requester* has.

I asked @gleemie, who is faculty at UCSD and creator of Turkopticon, her opinion. She said:
Discrimination by some definition — detecting and producing a difference within a general population — is literally what a lot of quantitative (and sometimes qualitative) analysis of populations does. The IRB is supposed to hold researchers responsible so they don’t make generalizations about certain categories of people (race, ethnicity, etc). Your question raises the question: 
 - how to make enable researchers to select for the populations they have IRB approval to research (which is different from letting them select anything they want because they think studying and publishing stats on Indian turkers would be fun)
 - how to ensure researchers are not filtering out certain workers for jobs except in ways that are sanctioned by IRB / scientific knowledge production standards

Building all that information into the profile employers have access to seems like a recipe for discrimination (including race, gender, age). To me, you’d want a system that would only let researchers filter for, say, 50-80 year olds, if they could present scientific justification (IRB protocol). Maybe daemo would have workers list this information but it would be listed privately in the system rather than being a default on which employers could filter. One way to do this would be to add a moderating layer that fields these discrimination requests. Another way would be to let employers filter but make their study publicly list the filters in operation and provide open data about employers’ discrimination choices so workers could keep an eye on it and ask that something be done about it. Another way to think about it would be that instead of employers paying the individual, they pay the whole pool — so all workers that make a commitment to be available actually get the payout, instead of the workers who are individually chosen. This would make a discrimination risk pool / coop. Maybe there are other ideas too….

dmorina:
build seems to be an issue, will update you when we identify the problem

shirishgoyal:
@dmorina checking issue with PR and build

shirishgoyal:
PR can be closed as extended by #666

dmorina:
closing see new PR extending this

PR #649, opened 2016-01-26
Issue timer

### LINKED ISSUE #645: timer on task submission screen
add a timer to the task submission screen which is based on the `timeout` field of the Project Model and it is a calculation of: now() -  taskworker created timestamp  should not greater than the timeout value, we can start with a default of 4hrs

sehgalvibhor:
Timer on the Task submission screen. In the format HH:MM:SS
screenshot1

After the default time of 4 hours, the time left shows 'Task expired'.
screenshot2

Thanks to @AKSHANSH47 for helping me out.
Closes #645


crowdsourcing/viewsets/task.py
```diff
@@ -59,9 +61,15 @@ def retrieve_with_data(self, request, *args, **kwargs):
         requester_alias = task.project.owner.alias
         project = task.project.id
         target = task.project.owner.profile.id
+        timeout = task.project.timeout
+        worker_timestamp = task_worker.created_timestamp
+        now = datetime.datetime.utcnow().replace(tzinfo=utc)
+        timeout = 4 * 3600  # Default value in Seconds
```
dmorina: assume this is in minutes and task.project.timeout will return you a value 
i.e `timeout = models.Integerfield(default=240)` and then you can update line 67 here

dmorina:
looks great, check that minor comment i left on the timeout, rest looks good to me :)

static/js/task/controllers/task.controller.js
```diff
@@ -58,12 +56,41 @@
...
                         });
                     }
+                    countdown("countdown1",0,self.time_left);
                 },
                 function error(data) {
                     $mdToast.showSimple('Could not get task with data.');
                 });
         }
 
+        function countdown(elementName, minutes, seconds )
```
shirishgoyal: Can you convert this into angular directive instead? Avoid manipulating DOM directly from controller.

sehgalvibhor:
Updated.

dmorina:
can you move the timer to a directive, for reuse later?

sehgalvibhor:
Moved to Directives.

static/templates/task/base.html
```diff
@@ -2,7 +2,7 @@
   <div class="_main-content-body">
     <div class="md-whiteframe-z1 md-padding" >
       <h3>Task #{{ task.taskData.id }} ({{ task.taskData.project_data.name }})</h3>
-
+      <div timer-directive my-text ='task.time_left'></div>
```
shirishgoyal: Can you change it to better names like 
a) timer-counter instead of timer-directive
b) seconds-left instead of my-text

Please validate value passed in my-text

shirishgoyal:
Can be closed as extended in #664

PR #654, opened 2016-01-31
Auto-accept / skip task

jamesalcatraz:
The back-end code for setting an auto skip feature has been added. 

1. Added a boolean field 'auto-accept' to user-preferences(default is false).
2. Initialized user preferences on registration.
3. Added methods to create, update and get user preferences.
4. User will see only 1 task per batch after submission if 'auto-skip' is disabled and he will return to the task-feed. Else, he will move on to the next task in the batch and continue so until all are either done or saved.


crowdsourcing/models.py
```diff
@@ -442,6 +442,7 @@ class UserPreferences(models.Model):
     currency = models.ForeignKey(Currency)
     login_alerts = models.SmallIntegerField(default=0)
     last_updated = models.DateTimeField(auto_now_add=False, auto_now=True)
+    auto_skip = models.BooleanField(default=True)
```
dmorina: probably auto_accept or accept_next is the right property here; should match the functionality

crowdsourcing/viewsets/user.py
```diff
@@ -151,7 +151,28 @@ class UserPreferencesViewSet(mixins.RetrieveModelMixin, mixins.UpdateModelMixin,
     queryset = UserPreferences.objects.all()
     permission_classes = [IsAuthenticated]
 
+    def create(self, request, *args, **kwargs):
```
dmorina: i guess we dont want to let them create preferences only update

crowdsourcing/models.py
```diff
@@ -442,6 +442,7 @@ class UserPreferences(models.Model):
     currency = models.ForeignKey(Currency)
     login_alerts = models.SmallIntegerField(default=0)
     last_updated = models.DateTimeField(auto_now_add=False, auto_now=True)
+    auto_accept = models.BooleanField(default=True)
```
dmorina: default false, so that people dont get annoyed I suppose

crowdsourcing/viewsets/user.py
```diff
@@ -155,3 +155,17 @@ def retrieve(self, request, *args, **kwargs):
         user = get_object_or_404(self.queryset, user=request.user)
         serializer = UserPreferencesSerializer(instance=user)
         return Response(serializer.data)
+
+    def update(self, request, user__username=None):
```
dmorina: lookup_field = 'user__username' and lookup_value_regex = '[^/]+'  
and then you can drop user__username

dmorina:
in general looking good, few minor things left comments

shirishgoyal:
@jamesalcatraz can you make frontend changes too to handle this?

shirishgoyal:
Can be closed as extended further in #669

PR #655, opened 2016-02-01
Mturk v2

dmorina:
changes to the logic of syncing:
releases only one assignment per task at a time to MTurk, when they submit a hit a new assignment is released 

@shirishgoyal @aginzberg @mbernst


mturk/interface.py
```diff
@@ -127,7 +126,7 @@ def set_notification(self, hit_type_id):
 
     def approve_assignment(self, task_worker):
         task_worker_obj = TaskWorker.objects.get(id=task_worker['id'])
-        if hasattr(task_worker_obj, 'mturk_assignments') and task_worker_obj.mturk_assignments is not None:
+        if hasattr(task_worker_obj, 'mturk_assignments') and task_worker_obj.mturk_assignments.first() is not None:
```
dmorina: I could change this to OneToOneField

PR #653, opened 2016-01-30
#647 Extend data source on authoring page

Nalinc:
Closes #647.

**Change summary:**
- Allow requesters to use static text with dynamic column names from attached csv file.
- Allow multiple occurrences of the same dynamic field.

Changes are available on https://daemo-develop.herokuapp.com/
Use these credentials to test [user@daemo.com/daemocrowd]

**General use-case(s):**
Question String: "hello {name} welcome to {country}"

 ```
data_sources=[{
     type:'static',
     value:'hello',
     position:0
},{
     type:'dynamic',
     value:'name',
     position:1
},{
     type:'static',
     value:' welcome to ',
     position:2
},{
     type:'dynamic',
     value:'country',
     position:3
}]
 ```
**Screenshots:**
rsz_1

rsz_12

PR #657, opened 2016-02-06
Task authoring auto complete

### LINKED ISSUE #656: Add Auto-complete in addition to Link Column for Data Sources and Options
image

a) Auto-complete should show all data columns available
b) Column added to data source / options will be available for use multiple times within same field

Nalinc:
Closes #656.

**Note:** This Pull Request is build upon changes from PR #653 which allows requester to include both static as well as dynamic tokens while task authoring.

**Change summary:**
Provide intuitive suggestions to requester while task-authoring, by showing the list of available column names from attached csv.

Changes are available at https://daemo-develop.herokuapp.com
Use the following credentials for quick lookup [user@daemo.com/daemocrowd]

**Use-Cases:**
Covers both the cases discussed in issue #656.
- allows requester to use the 'dynamic column names' as 'static' tokens, unless explicitly linked. Eg: Despite of having a column named 'frequency' in attached csv file, requester can use the word 'frequency'  in static string unless explicitly linked by the user.
- if csv is not attached or removed, this feature is simply ignored!

Auto-suggest dropdown appears when
- requester hits a specific character `{` ( this is similar to what we see in facebook/github while trying to tag a friend with @ symbol)
- requester types initial few characters of the "dynamic column string"

**Screenshot**
output_atszow


crowdsourcing/serializers/task.py
```diff
@@ -240,13 +240,19 @@ def get_template(self, obj, return_type='full'):
...
             if 'question' in aux_attrib and 'data_source' in aux_attrib['question'] and \
-                    aux_attrib['question']['data_source'] is not None and \
-                    aux_attrib['question']['data_source'] in data:
-                aux_attrib['question']['value'] = data[aux_attrib['question']['data_source']]
+                    aux_attrib['question']['data_source'] is not None:
+                    for data_source in aux_attrib['question']['data_source']:
+                        if 'value' in data_source and data_source['value'] is not None and \
+                                data_source['value'] in data:
+                            aux_attrib['question']['value'] = aux_attrib['question']['value']\
+                                .replace('{' + str(data_source['value']) + '}', str(data[data_source['value']]))
```
shirishgoyal: Should be able to handle any spaces within delimiters.

static/js/template/controllers/template.controller.js
```diff
@@ -235,18 +235,20 @@
...
+                    }).indexOf(data_source);
+        }
         function setDataSource(item, data_source){
-            if(!item.data_source || item.data_source != data_source){
-                item.data_source = data_source;
-                if(item.hasOwnProperty('value')) item.value = null;
-                if(item.hasOwnProperty('src')) item.src = null;
-                item.placeholder = 'will be filled from {' + data_source + '}';
+            //See if the data_source has already been linked in question text
+            if(item.value.search("{"+data_source+"}") > -1){
```
shirishgoyal: Should be able to handle any spaces within delimiters.

Nalinc:
@shirishgoyal: done with handling the extra spaces within delimiters`{ }`!

Following scenarios are covered:
- leading spaces { _ _dynamic token}
- trailing spaces {dynamic token _ _ }
- both leading and trailing spaces { _ _ dynamic token _ _ }
- redundant spaces in-between dynamic tokens {dynamic_ _ token}
- combination of above { _ _ dynamic _ _ token _ _ }

shirishgoyal:
Bugs:
a) Auto-complete doesn't work for options in Image, Audio, Remote widgets
b) Auto-complete doesn't work if Column name has spaces. After space, it doesn't show auto-complete menu.

Suggested Improvements:
a) Limit number of tags shown in auto-complete. Use scrollbar.

Nalinc:
Covered following scenarios:
- Extend data_sources for image, audio and iframe [check]
- Fix Auto-complete for options in Image, Audio, Remote widgets [check]
- Keep showing auto-complete if Column name has spaces. [check]
- Limit number of tags shown in auto-complete to 5 [check]
- Trim extra spaces between 'static' and '{column name}' before and after [check]

Though I've tried most of the edge-cases, I might have skipped few.

dmorina:
if we are doing autocomplete should we use angular-material autocomplete instead of `jquery.textcomplete`? @shirishgoyal thoughts

Nalinc:
@dmorina, @shirishgoyal: I investigated md-autocomplete few days back but had to drop it because of following reasons:
- It doesn't allow us to leverage regex as freely as textcomplete does. Like we can show the list of suggestions on hitting special characters `{` which may/may not be part of original pattern. Notice how I had to use two separate regex in current implementation(template.directive.js, Line:74,91).
- I found few issues while trying to reference the same word multiple times or even use different word again in the same textbox. It treats the input box as a stream of contiunous charcters and doesnt differentiate individual tokens.
- Doesnt seem to support padding of matched pattern with delimiters(`{`). 
- autocomplete box spans the width of entire input textbox, unlike textcomplete which takes only as much space as needed by the longest pattern.

In short, while its still possible to use md-autocomplete, but we might have to compromise with some features.
PS: textcomplete is available under MIT Licence which seem to fit well with Daemo.

shirishgoyal:
I agree with @Nalinc . I am fine with usage of jquery.textcomplete if it satisfies what we need. Many of the md components are still in development and may take time to build all those features and using them and extending them might be much more harder than integrating jquery based components with angular.

dmorina:
lgtm

dmorina:
resolve merge conflicts, apart from issue above lgtm

Nalinc:
fixed merge conflicts

dmorina:
@Nalinc can you fix the issue mentioned above with removing files

dmorina:
extended in #674

Issue #634, opened 2016-01-17
Auto-focus 1st field in Sign in/Register Form/Forgot Password Form


musicakc:
On signing in or registration, we can currently click on either of the options and the respective form is loaded for filling details. But the cursor doesn't automatically go to the first form field for you to start filling in. You have to first scroll down and then fill the details which should be automatic when you click on sign in/register. 
This would help in improving the user experience.

dmorina:
agreed, these pages need a redesign and this issue should be addressed

dmorina:
rename to 'Auto focus ..', maybe better

shirishgoyal:
Completed in #650

shirishgoyal:
Resolved in Resolve bugs #718

PR #659, opened 2016-02-08
Added autofocus for first fields of login and register templates

insomniac12:
Pull request for  #634


dmorina:
resolve conflicts please, lgtm otherwise

shirishgoyal:
This commit won't solve the problem completely as it requires directive approach for Angular. It will only work when page is loaded initially on login/register URLs.

dmorina:
agreed

dmorina:
@insomniac12 you may close this

Issue #660, opened 2016-02-10
boomerang permissions


dmorina:
-add permissions to API so that workers who are not yet qualified to do a task cannot see it when retrieving using IDs

shirishgoyal:
Implemented via AMT and workers have no direct access to platform

PR #661, opened 2016-02-15
#660 boomerang permissions

beingcooper:


crowdsourcing/viewsets/project.py
```diff
@@ -149,8 +149,26 @@ def list_feed(self, request, **kwargs):
                                                        'allow_feedback', 'price', 'task_time', 'owner',
                                                        'requester_rating', 'raw_rating', 'is_prototype',),
                                                context={'request': request})
+
         return Response(data=project_serializer.data, status=status.HTTP_200_OK)
 
+    @list_route(methods=['get'])
+    def task_check(self, request, **kwargs):
```
dmorina: should be a permission

static/js/crowdsource.directives.js
```diff
@@ -96,5 +97,17 @@
                 });
             }
         };
+    }   
+
+    function autoFocus($timeout) {
```
dmorina: why do we need this?

beingcooper: autoFocus is actually for issue #634.Not related to #660.

shirishgoyal: @dmorina Auto-focus needs a directive as single page apps do not support auto-focus when shifting to different views as page is not loaded again.

dmorina: okie, thanks

static/templates/project/prototype.html
```diff
@@ -1,38 +1,44 @@
...
+                <div class="form-content" layout="row">
+                    <md-input-container class="input-container" layout-align="center">
+                        <label>Tasks</label>
+
+                        <input class="input-content" name="num_tasks" required type="number" min="1"
+                               max="{{ max_rows }}" ng-model="num_rows">
+
+                    </md-input-container>
+                    <md-input-container class="input-container _ml-24" layout-align="center">
+                        <label>Users</label>
```
dmorina: since you are at it could you please rename this to Workers/task same as in the main Authoring screen, thanks

shirishgoyal: btw do we really need a dialog to popup ? :)

dmorina: Yes, we need this confirmation step for the prototype; we were considering different approaches sometime ago and we decided for dialog

shirishgoyal: Added this change in #665

crowdsourcing/permissions/project.py
```diff
@@ -13,3 +14,19 @@ def has_object_permission(self, request, view, obj):
         if request.method in permissions.SAFE_METHODS:
             return True
         return obj.worker.profile.user == request.user
+
+
+class IsProjectAvailable(permissions.BasePermission):
+    def has_object_permission(self, request, view, obj):
+        query = '''
+                SELECT wr.requester_id as id FROM get_worker_ratings(%(worker_profile)s) wr
```
dmorina: it seems a bit not efficient to fetch all ratings and then pick one, maybe another function which does a more specific lookup could be more efficient? i.e only get the ratings by this requester for this worker, same goes for project min_ratings

dmorina:
sitting for too long, will be included elsewhere

PR #662, opened 2016-02-16
Enable CORS for External Work Submission

shirishgoyal:
- Allow any server to post work via /api/done/
- Only allow JSON requests for external work. JSON is used to ensure request is pre-flighted one always and thus follow security measures for AJAX requests and CORS

Reference: https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS


dmorina:
do form posts work?

shirishgoyal:
Reverted JSON parser to allow both form POSTs and JSON

dmorina:
lgtm

PR #666, opened 2016-02-18
User Profile

shirishgoyal:
image

image

Pending City and Country Migrations


shirishgoyal:
@mbernst Do you have list of cities and countries we can allow at this stage? In other case, we can also use https://developers.google.com/places/javascript/ to fetch all information related to address.

mbernst:
I agree, let's go for any standardized set of places. Google's set — maybe
even using their widgets — doesn't seem like a terrible idea.

thunderboltsid:
Google's standard set looks good. Can you user their dataset without using their map visualisation? Or is the map something we want?

shirishgoyal:
@thunderboltsid I am specifically referring to this demo : https://developers.google.com/maps/documentation/javascript/examples/places-autocomplete-addressform

But after more discussion, data migration makes more sense I think since we might be linking city and country at other places too in future.

dmorina:
@thunderboltsid @shirishgoyal how are we doing on this one?

qwertyone:
Bad option: So I found a data set that is updated weekly, but it's a bad idea since the raw files are extremely large. ftp://ftp.nga.mil/pub2/gns_data/ this is posted to enable comparison of alternatives, and this is the worst.

shirishgoyal:
For now, have integrated google places API to fill in address. 
City / Country are not linked.

dmorina:
move view/edit profile into the More_Vert drop down

dmorina:
some of the drop downs are too wide e.g gender and ethnicity

crowdsourcing/viewsets/user.py
```diff
@@ -11,10 +11,11 @@
 from crowdsourcing.redis import RedisProvider
 from crowdsourcing.serializers.user import UserProfileSerializer, UserSerializer, UserPreferencesSerializer
 from crowdsourcing.permissions.user import CanCreateAccount
+from crowdsourcing.serializers.utils import CountrySerializer, CitySerializer
 from crowdsourcing.utils import get_model_or_none
 
 
-class UserViewSet(mixins.RetrieveModelMixin, mixins.CreateModelMixin, viewsets.GenericViewSet):
+class UserViewSet(viewsets.ModelViewSet):
```
dmorina: i dont think we wanna allow all 5 methods by default, just add mixins.UpdateModelMixin

dmorina:
screen shot 2016-04-27 at 3 13 37 pm
could we have no defaults here?

otherwise lgtm, see comment

PR #668, opened 2016-03-01
Angular-UI-Router

shirishgoyal:


dmorina:
screen shot 2016-03-02 at 10 29 54 am

static/js/crowdsource.routes.js
```diff
@@ -2,155 +2,323 @@
...
+                views: {
+                    'navbar': navbar,
+                    'content': projectReview
+                },
+                authenticate: true,
                 resolve: {
-                    resolvedData: function ($route, Project) {
-                        return Project.retrieve($route.current.params.projectId);
+                    resolvedData: function ($stateParams, Project) {
+                        return Project.retrieve($stateParams.params.projectId);
```
dmorina: i think this is failing

dmorina:
lgtm overall, thanks for this
a comment left above

dmorina:
btw I had a discussion with Christopher some time ago and an idea was to keep 960px for bigger screens and min-width: 768px; so that we have a bit more space

shirishgoyal:
I would rather just use google specs as defined here : https://material-design.storage.googleapis.com/publish/material_v_4/material_ext_publish/0B8olV15J7abPSGFxemFiQVRtb1k/layout_adaptive_breakpoints_01.png

https://www.google.com/design/spec/layout/responsive-ui.html#responsive-ui-breakpoints

shirishgoyal:
Fixed bug in project review

dmorina:
lgtm

PR #669, opened 2016-03-06
Auto-accept next task

shirishgoyal:
- Extend #654 with frontend

image


crowdsourcing/viewsets/task.py
```diff
@@ -227,6 +236,13 @@ def submit_results(self, request, *args, **kwargs):
             task_worker.task_status = task_status
             task_worker.save()
 
+            try:
+                user_preferences, created = models.UserPreferences.objects.get_or_create(user=request.user)
+                user_preferences.auto_accept = auto_accept
+                user_preferences.save()
```
dmorina: let's change this only when we need to, meaning only when the checkbox state is changed

crowdsourcing/serializers/user.py
```diff
@@ -137,9 +164,17 @@ def create(self, **kwargs):
...
         user.last_name = self.validated_data['last_name']
         user.save()
+
         user_profile = models.UserProfile()
         user_profile.user = user
         user_profile.save()
+
+        user_preferences = models.UserPreferences()
+        user_preferences.user = user
+        user_preferences.currency = models.Currency.objects.create()
```
dmorina: probably only system can create currencies

static/templates/task/base.html
```diff
@@ -12,6 +11,11 @@
                     <div md-template-compiler="item" editor="false" instance="template"></div>
                 </div>
                 <div layout="row" class="task-actions" layout-align="end center">
+                    <md-checkbox name="auto-accept" ng-model="task.auto_accept" aria-label="Auto-accept next task"
```
dmorina: should we make it a switch since there is only one option?

crowdsourcing/viewsets/task.py
```diff
@@ -72,10 +71,19 @@ def retrieve_with_data(self, request, *args, **kwargs):
         worker_timestamp = task_worker.created_timestamp
         now = datetime.datetime.utcnow().replace(tzinfo=utc)
         time_left = int((timeout * 60) - (now - worker_timestamp).total_seconds())
+
+        auto_accept = False
+        try:
+            user_prefs = UserPreferences.objects.get(user=request.user)
```
dmorina: either use something like request.user.preferences.auto_accept or use get_model_or_none which wraps this for you

crowdsourcing/viewsets/task.py
```diff
@@ -227,6 +236,15 @@ def submit_results(self, request, *args, **kwargs):
             task_worker.task_status = task_status
             task_worker.save()
 
+            try:
+                user_preferences, created = models.UserPreferences.objects.get_or_create(user=request.user)
```
dmorina: do not update user preferences in this viewset, use the userpreferences viewset for that

static/templates/task/base.html
```diff
@@ -12,6 +11,10 @@
                     <div md-template-compiler="item" editor="false" instance="template"></div>
                 </div>
                 <div layout="row" class="task-actions" layout-align="end center">
+                    <md-switch name="auto-accept" ng-model="task.auto_accept" aria-label="Auto-accept next task">
```
dmorina: onChange of this field do a PUT to UserPreferencesViewSet

dmorina:
lgtm

PR #672, opened 2016-03-15
Mturk v3

dmorina:
-allows requesters to use their own AWS keys
-changes to mturk sync


mturk/tasks.py
```diff
@@ -1,28 +1,41 @@
...
+from django.db.models import Q
+from django.contrib.auth.models import User
+from crowdsourcing.crypto import AESUtil
 
 
 @celery_app.task
 def mturk_publish():
-    provider = MTurkProvider(SITE_HOST)
-    projects = Project.objects.filter(deleted=False, min_rating__lt=0.61)
+    projects = Project.objects.filter(~Q(owner__profile__user__mturk_account=None), deleted=False, min_rating__lt=0.61,
```
shirishgoyal: Should this min rating value be global setting? OR is it computed ?

static/js/project/controllers/project.controller.js
```diff
@@ -47,10 +49,41 @@
...
+            ).finally(function () {
+
+            });
+        }
+
+        function create_or_update_aws() {
+            if (self.aws_account.client_secret == null || self.aws_account.client_id == null) {
+                $mdToast.showSimple('Client key and secret are required');
+            }
+            User.create_or_update_aws(self.aws_account).then(
```
shirishgoyal: This should run when values are provided. 

Is there a way we can validate if client key and secret are valid?  
If no, we should handle wrong key and secret.

mturk/tasks.py
dmorina: this is a global value anything greater than 0 and lower than 1 works (magic random choice to puzzle people lol)

static/js/project/controllers/project.controller.js
dmorina: we could do a test authentication against AWS, sure

dmorina: @shirishgoyal done

mturk/tasks.py
shirishgoyal: Let's shift it to settings then :)

static/templates/project/authoring.html
```diff
@@ -99,17 +99,26 @@
...
+                            <label>Time Allotted (in min.)</label>
+                            <input name="timeout" type="number" ng-model="project.project.timeout" min="1">
+                        </md-input-container>
                     </div>
-                    <md-input-container class="md-block larger">
-                        <label>Time Allotted (in min.)</label>
-                        <input name="timeout" type="number" ng-model="project.project.timeout" min="1">
-                    </md-input-container>
+                    <div>
+                        <div class="_sub-header">Post my task to Mechanical Turk if nobody on Daemo is available</div>
```
shirishgoyal: We may not need this header at all. There is no need to disclose when MTurk will be used.

static/templates/project/authoring.html
```diff
@@ -99,17 +99,26 @@
...
+                        <md-switch ng-change="project.AWSChanged($event)" ng-model="project.project.post_mturk">Post to Mechanical Turk</md-switch>
+                        <div class="_small-note _error">{{ project.AWSError }}</div>
```
shirishgoyal: I think better to use this image to avoid confusion on which MTurk we are referring to.
https://images-na.ssl-images-amazon.com/images/G/01/webservices/mechanical-turk/logoAI3.gif

static/templates/project/authoring.html
dmorina: i think we need too let them know (since they will use AWS with MTurk policies enabled)
wording might change, this was from Michael for the moment

PR #674, opened 2016-03-17
Auto-complete in Task Authoring

shirishgoyal:
Extends PR #657 

- Fixes CSS issues
- Fixes updating dynamic sources when file is removed or re-added.


dmorina:
lgtm

PR #681, opened 2016-04-29
Task and project expiration

aginzberg:
Implements expiration for tasks and projects. If a worker holds a task longer than the timeout, he relinquishes it back to the platform and when a project is past its deadline it no longer appears on the taskfeed.


crowdsourcing/viewsets/project.py
```diff
@@ -158,7 +158,8 @@ def list_feed(self, request, **kwargs):
                                                        'allow_feedback', 'price', 'task_time', 'owner',
                                                        'requester_rating', 'raw_rating', 'is_prototype',),
                                                context={'request': request})
-        return Response(data=project_serializer.data, status=status.HTTP_200_OK)
+        projects_filtered = filter(lambda x: x['available_tasks'] > 0, project_serializer.data)
```
dmorina: better than what we have now :D, eventually we will do this in sql

dmorina:
lgtm

PR #689, opened 2016-05-12
Qualifications

dmorina:
- adds back end for qualifications and global blacklists to projects


shirishgoyal:
1) When there is no project and user clicks on Project Preferences, it errors out

image

2) The preferences dialog doesn't allow scroll

crowdsourcing/models.py
```diff
@@ -400,22 +406,22 @@ class ActivityLog(TimeStampable):
...
     type = models.IntegerField(choices=TYPE, default=TYPE_STRICT)
 
 
 class QualificationItem(TimeStampable):
-    qualification = models.ForeignKey(Qualification, related_name='items')
-    attribute = models.CharField(max_length=128)
-    operator = models.CharField(max_length=128)
-    value1 = models.CharField(max_length=128)
-    value2 = models.CharField(max_length=128)
+    qualification = models.ForeignKey(Qualification, on_delete=models.CASCADE)
```
shirishgoyal: Please add related_name='items'

crowdsourcing/models.py
```diff
@@ -400,22 +406,22 @@ class ActivityLog(TimeStampable):
...
+    expression = JSONField()
+    next_op = models.CharField(max_length=16, default='empty')
```
shirishgoyal: Is this being used ?

dmorina:
@shirishgoyal fixed, removed and updated

crowdsourcing/viewsets/task.py
```diff
@@ -129,6 +130,7 @@ def create(self, request, *args, **kwargs):
         serialized_data = {}
         if http_status == 200:
             serialized_data = TaskWorkerSerializer(instance=instance).data
+            update_worker_cache.delay([instance.worker_id], 'CREATED')
```
shirishgoyal: Can you avoid using strings (like 'CREATED' here) at multiple places and define a type rather?

shirishgoyal:
image

Doesn't allow choosing 0

Should possibly allow >=, <=, = as well

crowdsourcing/viewsets/task.py
dmorina: we can group them together somewhere

dmorina:
0 shouldnt be allowed > or < than 0, doesnt make sense
initially these two should be sufficient

shirishgoyal:
image

Doesn't show anywhere if it is an AND condition 

If it is AND, the same attribute should follow the validation conditions set previously
Like here if num completed tasks is set to greater than 10, then it should not allow setting it again as less than 10

PR #714, opened 2016-06-08
User profile demographics and caching

sharonzhou:
Added education level and income
Added caching of user profile elements: birthday year, ethnicity, gender

PR #716, opened 2016-06-09
User profile demographics and caching ref #714

dmorina:
on behalf of Sharon
Added education level and income
Added caching of user profile elements: birthday year, ethnicity, gender


dmorina:
lgtm

PR #718, opened 2016-06-14
Resolve bugs

shirishgoyal:
Messaging
- Prevent multiple duplicate connections
- Prevent multiple duplicate conversations between same users
- UI issue with text overflow
- Handle error when no overlays
- Show error during message posting

Browser version compatibility notification
Google Maps API key error
Auto-focus on first field for Authentication forms


dmorina:
lgtm

PR #719, opened 2016-06-15
Revisions

dmorina:
-adding revisions to Project related models
-optimized feed query for `available_tasks`

todo: 
-add existing results to conflicts
-revert to previous revision when draft revision is discarded


shirishgoyal:
image

Substitution didn't work in this branch for data uploaded.

shirishgoyal:
image

a) It is not clear what X means here and what is it supposed to do. I assume it relaunches task but it is not clear from this.
b) Pagination arrows look disconnected from rest of the UI. Also pagination should be visible only if there are page full of records and it should allow back-forth only if records are available for that page
c) Does Relaunch-all ignore whatever action was taken on individual rows (pressing X)?

PR #722, opened 2016-06-28
Address bugs

nevenwt:
Adds city, country, and state to database.
Allows user to remove their address from their information
If a user chooses to add their address, it must contain a city and country at the least (address and street are optional)


static/js/user/controllers/user.controller.js
```diff
@@ -202,41 +212,98 @@
...
+                return;
             }
+            if (vm.addressSearchValue !== "" && vm.user.address_text.place_id !== undefined) {
+                var service = new google.maps.places.PlacesService(document.getElementById('node'));
+                service.getDetails({placeId:vm.user.address_text.place_id}, function(result, status) {
+                    console.log(result);
+                    var street_number = "";
+                    var street = "";
+                    user.location = {};
+                    for (var i = 0; i < result.address_components.length; i++) {
```
dmorina: use _.filter here 
and please resolve conflicts

dmorina:
lgtm

PR #729, opened 2016-07-08
MTurkNotification change, MTurk assignment status, components extension: patterns for "text" component and "slider" component

nevenwt:
Changed MTurkNotification to unpack the JSON data into a few fields (hit_id, hit_type_id, assignment_id, event_type). Fixed a bug with the creation of notifications.

Implemented various patterns for the text component for task authorship
-Plain text (contains, doesn't contain, email)
-Number (greater than, greater than equal to, less than, less than equal to, equal to, between)
-Regular Expression

Added "slider" component for task authorship


dmorina:
you merged with prev pr?

static/css/app.css
```diff
@@ -1,2 +1,2215 @@
...;display:-webkit-flex;display:-ms-flexbox;display:flex;-webkit-box-orient:vertical;-webkit-box-direction:normal;-webkit-flex-direction:column;-ms-flex-direction:column;flex-direction:column;padding-bottom:4px}._chat-overlay ._chat-body ._overlay-input-wrapper{font-size:14px;height:auto;position:relative}._chat-overlay ._chat-body ._overlay-input-wrapper md-input-container{margin:0;padding:0 1px 4px}._chat-overlay ._chat-body ._overlay-input-wrapper md-input-container .md-input{border-bottom-width:0;line-height:normal;width:254px;vertical-align:middle;font-family:Fira Sans,sans-serif;padding:6px}.md-button.md-fab.md-fab-chat-overlay{float:right;position:relative;top:20px;right:15px;z-index:2}.md-fab-chat-overlay:after{clear:both}._notification:not(._read):after{content:'';display:inline-block;width:6px;height:6px;border-radius:3px;background-color:#ff5c80;vertical-align:top;margin-left:1px}._online:not(._offline):before{background-color:#009688}._offline:not(._online):before,._online:not(._offline):before{content:'';display:inline-block;width:8px;height:8px;border-radius:8px;vertical-align:middle;margin-left:0;margin-right:4px}._offline:not(._online):before{background-color:#b71c1c}
```
dmorina: do not track .css*

static/js/template/services/template.service.js
```diff
@@ -80,17 +80,25 @@
                     icon: 'text_fields',
                     type: 'text',
                     sub_type: 'text',
+                    pattern_type: 'email',
```
dmorina: the properties on level 0 are the ones which every component has, so better to move this to `aux_attributes` as that's the object which has an unspecified structure and the property may or may not exist

shirishgoyal:
a) Fails on Item validation for normal text - Expecting a pattern 
TypeError at /api/task-worker-result/submit-results/
unhashable type: 'dict'

b) Allows incorrect regex input and complains in JS console
Allows any input for such regex field

c) When all fields are filled in with some invalid inputs, it still shows error as - All fields are required.

d) Slider component allows choosing max limit or min limit (when range is say 0-15) even when step is 2 while it doesn't allow choosing intermediate odd values.

nevenwt:
Fixed a, b, and e. Did not fix c because the behavior is okay. Did not fix d because it is simply how an angular slider is implemented, also does not seem like a problem.

shirishgoyal:
Validate for : No two fields should have the same field/key name.

PR #730, opened 2016-07-15
Dialogs

shirishgoyal:
- Show prototype dialog only for first publish
- Handle delete template item bug in task authoring (called Delete api twice)
- Updates to task retrieve api for bot scripts


static/js/project/controllers/project.controller.js
```diff
@@ -498,10 +502,12 @@
                         },
                         function error(response) {
                             console.log(response[1]);
+
                             if (response[1] == 402) {
-                                console.log('insufficient fnd');
+                                console.log('insufficient funds');
```
dmorina: :D

PR #823, opened 2016-10-21
Show correct response in task if gold_truth present

shirishgoyal:
If "gold_truth" field is present in task input, it is used to show correct response to worker after response is submitted.

All input fields must be present in gold_truth object named using the field name defined in the task design. Checkboxes should be represented as list of strings as correct options.

Example:
```json
{
    "id": 1,
    "url": "http://philamuseum.org/collections/permanent/85407.html?mulR=5090%7C1",
    "gold_truth": {
      "stance": "Anti-Legalization",
      "reason":["I enjoyed the topic","It was hilarious"],
      "repeat":"Yes",
      "title":"Dark Matter"
    }
}
```


mbernst:
Nice!

PR #835, opened 2017-02-24
New task creation flow (not fully done)

hashemelezabi:
To do:
- implement horizontal list of 'blocks' or 'cards' that each represent one type of task (eg. image tagging, etc.)
- implement arrow at the bottom of each tab that 'scrolls' to the next tab (ie. next step in task creation)
- make the tabs disappear and display a new view entirely when "Done" is pressed

Notes:
- I set the "md-dynamic-height" attribute to "true" because I had a problem with the size of the views under each tab (they were fixed to a small size), but I don't know if we want to keep it dynamically resizeable like that
- I didn't do the launch part yet (after pressing "Done")


dmorina:
closing as it has been covered previously in another PR

PR #867, opened 2017-06-15
Preview

dmorina:


dmorina:
@mbernst check and adjust the language if necessary 
` 46 static/templates/project/preview.html `

mbernst:
Will look! Need to reinstall dev environment on new laptop. Yayayayay

shirishgoyal:
@dmorina can you check the fix I pushed for height issue? it should not require opening new page

dmorina:
We got rid of it, now it opens a new tab

mbernst:
We need to decide if preview is task level (a la Mechanical Turk) or project level.

I'm seeing some weird behavior. I accept the project with 494 tasks. Then I press "skip" to skip the first task, and it returns everything and takes me back to the task feed. Shouldn't it take me to the second task in the project? Actually, the same thing happens even if I submit the task. I think this has to do with auto-accept?

dmorina:
correct, that's what auto accept does, if unchecked it will send you back to taskfeed

dmorina:
what does a project level preview mean? is that just the template showing the {{variables}}?
I thought we decided to render first row of the csv ?

mbernst:
Yeah it previews the project instead of each task. On AMT you preview each task before accepting, which is why autoaccept is necessary

mbernst:
Need to think about the right model...

dmorina:
Ok then, let me know

dmorina:
However, 4 of the workers who have been working on your tasks have auto-accept enabled so it doesn't seem to be an issue as it is right now, also there were no complaints from them about this.
My vote would be just keep it as it is until someone faces an issue.

static/templates/project/preview.html
```diff
@@ -0,0 +1,46 @@
...
+        <span class="_nav-title _truncated-text" style="max-width: 650px">{{ taskfeed.previewedProject.name }}</span>
+
+        <div class="_content-actions">
+
+
+        </div>
+    </div>
+    <div class=" _error-inset" style="margin: 16px 16px 16px 32px"
+         ng-if="project.project.status==project.status.STATUS_IN_PROGRESS || project.project.status==project.status.STATUS_PAUSED">
+        You are previewing this project. Click "Accept" to start working on it.
```
dmorina: @mbernst do you have a sentence for me here?

mbernst: That's fine for now. I need to figure out the broader situation of previewing...

mbernst: In the meantime lgtm

PR #880, opened 2017-06-22
Randoption

abhay-venkatesh:
Made shuffling of options possible on display to task worker. 

Removed uwsgi-dev.ini from repo and added to .gitignore

Added isEditPreview parameter to the task preview page before editing


dmorina:
looks good

PR #944, opened 2017-08-03
updating endpoints relevant to daemo-api-client

dmorina:
make api client great again


mbernst:
I fail to see what this has to do with melisandre

PR #959, opened 2017-08-26
Added form to new user handling in sign up

markwhiting:
Added a link to google form when `register.error && form.$pristine` are true for a users submission.

### DIFF
--- a/static/js/authentication/controllers/register.controller.js
+++ b/static/js/authentication/controllers/register.controller.js
@@ -26,6 +26,7 @@
                 self.addressSearchValue = null;
                 self.getAddress = getAddress;
                 self.errors = [];
+                self.registrationEnabled = true;
 
                 // var PlaceService = new google.maps.places.AutocompleteService();
 
@@ -41,6 +42,11 @@
                             $mdToast.showSimple('Email with an activation link has been sent.');
                             $state.go('auth.login');
                         }, function (data, status) {
+                            
+                            if (status === 403) {
+                                self.registrationEnabled = false;
+                                return;
+                            }
 
                             //Global errors
                             if (data.data.hasOwnProperty('detail')) {
--- a/static/js/authentication/controllers/register.controller.js
+++ b/static/js/authentication/controllers/register.controller.js
@@ -43,7 +43,7 @@
                             $state.go('auth.login');
                         }, function (data, status) {
                             
-                            if (status === 403) {
+                            if (data.status === 403) {
                                 self.registrationEnabled = false;
                                 return;
                             }
--- a/static/templates/authentication/register.html
+++ b/static/templates/authentication/register.html
@@ -74,7 +74,7 @@
         you agree to provide Daemo accurate and complete information about you and your business, and you authorize Daemo to share it and transaction
         information related to your use of the payment processing services provided by Stripe. Daemo does not store
         any banking or credit card information.</div-->
-    <div class="authentication-error" layout="row" ng-show="register.error=='We are currently in closed beta. If you\'d like an account, email daemo@cs.stanford.edu with a short description of what you\'d like to use Daemo for.' && form.$pristine">
+    <div class="authentication-error" layout="row" ng-show="registrationEnabled == false && form.$pristine">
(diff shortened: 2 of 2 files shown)


static/templates/authentication/register.html
```diff
@@ -74,8 +74,9 @@
         you agree to provide Daemo accurate and complete information about you and your business, and you authorize Daemo to share it and transaction
         information related to your use of the payment processing services provided by Stripe. Daemo does not store
         any banking or credit card information.</div-->
-    <div class="authentication-error" layout="row" ng-show="register.error && form.$pristine"
-         ng-bind="register.error"></div>
+    <div class="authentication-error" layout="row" ng-show="register.error && form.$pristine">
+      <span>We are currently in closed beta. Please <a href="https://goo.gl/forms/8eNjaR8pKc3aIpci1">fill out this form</a> to register interest in using Daemo.</span>
```
dmorina: register error can be anything here, can you check in angular if the response code is 403 or whatever you get when registration closed exception is raised or worst case you can do a string comparison of current error message

markwhiting:
@dmorina, I updated to include that check.

static/templates/authentication/register.html
```diff
@@ -74,8 +74,9 @@
         you agree to provide Daemo accurate and complete information about you and your business, and you authorize Daemo to share it and transaction
         information related to your use of the payment processing services provided by Stripe. Daemo does not store
         any banking or credit card information.</div-->
-    <div class="authentication-error" layout="row" ng-show="register.error && form.$pristine"
-         ng-bind="register.error"></div>
+    <div class="authentication-error" layout="row" ng-show="register.error=='We are currently in closed beta. If you\'d like an account, email daemo@cs.stanford.edu with a short description of what you\'d like to use Daemo for.' && form.$pristine">
```
dmorina: and what if it's something else? :P 
i'd recommend doing this in the controller

shirishgoyal: I think better to hide the signup button or even the form itself till registration is not open. Also the link should open on new page (target="_blank"). Can we change the text to something like -- fill the form to add yourself to the waitlist and we will reach out to you as soon as we have more slots open for you.

mbernst:
Don't we need the form open in order to enroll new requesters?

markwhiting:
I’d previously thought of pushing the sign up form to a secret link and just making the data collection form link directly from the ‘getting started’ link. 

This might be more tricky, but might also make things easier for users in the short term.

markwhiting:
@dmorina updated with those changes.

static/templates/authentication/register.html
```diff
@@ -74,8 +74,9 @@
         you agree to provide Daemo accurate and complete information about you and your business, and you authorize Daemo to share it and transaction
         information related to your use of the payment processing services provided by Stripe. Daemo does not store
         any banking or credit card information.</div-->
-    <div class="authentication-error" layout="row" ng-show="register.error && form.$pristine"
```
dmorina: Keep this around in case it's not a 403

markwhiting:
@dmorina, Ok, sure. I've added this and filtered out when it is a 403, to avoid 2 errors appearing.

Not sure if thats the right way to go about it, let me know what you think.

static/js/authentication/controllers/register.controller.js
```diff
@@ -41,6 +42,11 @@
                             $mdToast.showSimple('Email with an activation link has been sent.');
                             $state.go('auth.login');
                         }, function (data, status) {
+                            
+                            if (status === 403) {
```
dmorina: alright, status seems to be undefined, don't use that then just do `data.status`

static/templates/authentication/register.html
```diff
@@ -74,8 +74,10 @@
...
-         ng-bind="register.error"></div>
+    <div class="authentication-error" layout="row" ng-show="registrationEnabled == false && form.$pristine">
```
dmorina: registrationEnabled isn't in the scope and are you setting the form to $pristine if you are using it as a condition here?

PR #986, opened 2017-10-13
Making Tasks psots invisible

markwhiting:
This makes tasks posts invisible on the forum.


dmorina:
is this still relevant? has been sitting here for some time

markwhiting:
Yes. I've held on it because I expected that we might solve this problem differently when reconsidering prototype tasks, but it looks like the problem will remain.

PR #1020, opened 2017-11-29
Improved proposal template

markwhiting:
- Added breaking consensus instructions.
- Restructured to include steps for execution
- Improved links

### DIFF
--- a/issue_template.md
+++ b/issue_template.md
@@ -1,9 +1,11 @@
 ### The problem
-Replace this text with a clear statement of a problem including any relevant supporting files ([e.g.](https://github.com/crowdresearch/daemo/issues/1017)). If its a software abnormality, explain how it should act, how it does act, and how to reproduce it.
+1. Replace this text with a clear statement of a problem (e.g. #1017) including any relevant supporting files. If its a software abnormality, explain how it should act, how it does act, and how to reproduce it.
+2. Add a title in the field above.
+
+**To report a bug**, not a `proposal`, then simply complete the `problem` section and remove the rest of the template.
 
 ### The proposal
-Replace this text with a clear statement the proposed approach including any relevant supporting files ([e.g.](https://github.com/crowdresearch/daemo/issues/1017)).
-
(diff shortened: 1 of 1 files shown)


issue_template.md
```diff
@@ -1,9 +1,11 @@
...
-
-It can strengthen your proposal to include *pros* and *cons* and your thoughts about the *implications* (short and long term) of it being executed.
+**To report a bug**, not a `proposal`, then simply complete the `problem` section and remove the rest of the template.
 
-**Note:** if this is a bug, not a `proposal`, then simply complete the `problem` section and remove the rest of the template. 
+### The proposal
+3. Replace this text with a clear statement the proposed approach (e.g. #1017) including any relevant supporting files. It can strengthen your proposal to include *pros* and *cons* and your thoughts about the *implications* (short and long term) of it being executed.
+- - -
+#### Do not remove  
+**To break consensus** use [this template](https://gist.githubusercontent.com/markwhiting/ea8875602acd11e79002763735701503/raw/a8da103e23c07b74b4007459170104af04e38a33/break_consensus.md). You may also install this template to your account by following [these instructions](https://help.github.com/articles/creating-a-saved-reply/).
```
dmorina: this doesn't apply to most of the issues which are created by us regarding the daemo platform, so the existing version is the most generalizable  version which can be used for normal/traditional issues

markwhiting: You mean lines 9-11?

Do you think it would be better if it didn't ask not to be removed?

dmorina: i mean all the new text, yes everything below "Do not remove"
nope, consensus is not something relevant to the way we handle github issues at least the ones Daemo code related so that doesn't generalize to our use case

markwhiting: Sure. Do you suggest a better way to do this?

One option would be to take away the 'do not remove' part and leave the text below. That way issues that are unrelated to the consensus process would not be impacted because this would all be deleted anyway. 

Another option would be to reframe these instructions a little. 

The purpose of including them was just to name sure that people know how to break consensus in a reasonable way. 

What do you think?

dmorina:
ok, let's try this

PR #1040, opened 2018-01-03
Fixing the IRB to align with what we have on file at Stanford

markwhiting:


neilthemathguy (commented):
Can you move #1041 to the governance section https://github.com/crowdresearch/collective and link #1040 from there

dmorina (approved):
looks fine

shirishgoyal:
@markwhiting Please follow the defined governance process. All the proposals should be created on https://github.com/crowdresearch/collective and only when passed consensus, they should be copied to https://github.com/crowdresearch/daemo if related to implementation in daemo platform. Also please attach the IRB submitted to Stanford for reviewers to verify if changes made are in sync. Consider this issue open till these tasks are taken care of.

markwhiting:
Hey, this was treated as a bug because the state was inconsistent with the way it was designed or intended. In that sense theres nothing to have consensus on and there's no change associated with this other than to fix a mistake in this file. I am not really sure how else this could be considered but I'd be happy to talk about that at our next meeting if you'd like.

Here's the waiver on file with Stanford: [38845 Waiver.pdf](https://github.com/crowdresearch/daemo/files/1610758/38845.Waiver.pdf)

shirishgoyal:
> Bug fixes occurring under blanket approvals still require proposals that stay open for the weekly cycle: the goal is to achieve consensus on whether any larger systemic changes need to be made to address it in the future.

